PR #78, opened 2026-07-29
refactor: extract terminal emulator behind a backend trait

aymanbagabas:
Puts the terminal emulator behind an `Emulator` trait so additional backends (libghostty, xterm.js) can be added without touching the daemon, render, or assert layers, then makes the neutral grid vocabulary lossless enough for those backends to agree on.

**The seam**

- `Emu` → `AlacrittyEmu`, now the first `impl Emulator`; neutral grid types moved to `terminal::cell`
- `CommandTracker` lifted out of the emulator into `TermState` — OSC 133/633/7 parsing is backend-neutral, so every backend gets identical shell integration by construction rather than reimplementing it
- adds `emulator_conformance_tests!`, backend-agnostic tests any new backend has to pass

**The cell model**

The vocabulary flattened away detail the emulator had already parsed, and each loss surfaced downstream:

- `Color::Default` was indistinguishable from ANSI 0, so `expect --fg 0` matched a default-colored cell and reported it as black when the theme paints it light gray. Colors are `Option<Color>`; a default cell matches nothing.
- A blank cell and the second column of a double-width character were both `""`, and text extraction substituted a space for each — inventing a column, so `你a` extracted as `你 a` and shifted every column after it. Blanks hold a real space, `""` means continuation, extraction skips it.
- Five underline styles collapsed into one bool and the underline color was dropped. `Option<Underline>` keeps both.
- Palette slots 0-15 are themeable and 16-255 are fixed, but both landed in `Color::Idx`. Now `Color::Named` and `Color::Idx`, split by numeric range rather than by how the escape sequence spelled it, so a backend that only carries a palette index agrees with one that carries a name.

Attributes move into a `bitflags` set, which adds blink and reduces the per-cell style comparison in the render and monitor loops to one integer compare. `ch` becomes a `CompactString`, inline up to 24 bytes, so the grid stops heap-allocating per cell and can hold combining marks the backend was discarding.

**Wire format**

Colors are unchanged: named and indexed both serialize to their palette index. `cells` was reporting four of seven attributes, so `dim`, `invisible`, `strike` and `blink` are added along with `underline_style` and `underline_color`. `blink` is always false from alacritty, which parses SGR 5/6/25 and discards them, but ghostty and xterm.js both track it. A continuation cell now reports `""` rather than a space; the binding types document it.

### DIFF
--- a/bindings/js/src/types.ts
+++ b/bindings/js/src/types.ts
@@ -1,5 +1,12 @@
 export type Color = "default" | number | string;
 
+export type UnderlineStyle =
+  | "single"
+  | "double"
+  | "curly"
+  | "dotted"
+  | "dashed";
+
 export type Shell =
   | "bash"
   | "powershell"
@@ -29,9 +36,17 @@ export interface Cell {
   fg: Color;
   bg: Color;
   bold: boolean;
+  dim: boolean;
   italic: boolean;
-  underline: boolean;
   inverse: boolean;
+  invisible: boolean;
+  strike: boolean;
+  /** Always `false` from the alacritty backend, which cannot report blink. */
+  blink: boolean;
+  underline: boolean;
+  underline_style: UnderlineStyle | null;
+  /** `null` means the underline follows the text color. */
+  underline_color: Color | null;
 }
 
 export interface State {
--- a/bindings/js/src/types.ts
+++ b/bindings/js/src/types.ts
@@ -24,6 +24,7 @@ export interface Size {
 export interface Cell {
   x: number;
   y: number;
+  /** The cell's grapheme; `" "` when blank, `""` for the second column of a double-width character. */
   char: string;
   fg: Color;
   bg: Color;
--- a/bindings/js/src/types.ts
+++ b/bindings/js/src/types.ts
@@ -1,6 +1,8 @@
 export type Color = "default" | number | string;
 
+/** `"none"` is a value, not an absence: an un-underlined cell reports it. */
 export type UnderlineStyle =
+  | "none"
   | "single"
   | "double"
   | "curly"
@@ -43,10 +45,11 @@ export interface Cell {
   strike: boolean;
   /** Always `false` from the alacritty backend, which cannot report blink. */
   blink: boolean;
+  /** Shorthand for `underline_style !== "none"`. */
   underline: boolean;
-  underline_style: UnderlineStyle | null;
-  /** `null` means the underline follows the text color. */
-  underline_color: Color | null;
+  underline_style: UnderlineStyle;
+  /** `"default"` means the underline follows the text color. */
+  underline_color: Color;
 }
 
 export interface State {
--- a/bindings/js/src/types.ts
+++ b/bindings/js/src/types.ts
@@ -48,7 +48,11 @@ export interface Cell {
   /** Shorthand for `underline_style !== "none"`. */
   underline: boolean;
   underline_style: UnderlineStyle;
-  /** `"default"` means the underline follows the text color. */
+  /**
+   * `"default"` means the underline follows the text color. Tracked
+   * independently of `underline_style`, so a cell that set SGR 58 without an
+   * underline still reports the color it would use.
+   */
   underline_color: Color;
 }
 
--- a/src/assert/color.rs
+++ b/src/assert/color.rs
@@ -53,43 +53,39 @@ fn parse_hex(hex: &str) -> anyhow::Result<(u8, u8, u8)> {
 }
 
 /// Does a cell's resolved color match the expected color?
-pub fn matches(cell: Color, expected: &Expected) -> bool {
+///
+/// A cell using the terminal default never matches: the default is whatever
+/// the viewer's theme picks, so no concrete value can be claimed for it.
+pub fn matches(cell: Option<Color>, expected: &Expected) -> bool {
+    let Some(cell) = cell else {
+        return false;
+    };
     match expected {
-        Expected::Ansi256(n) => match cell {
-            Color::Default => *n == 0,
-            Color::Idx(i) => i == *n,
-            Color::Rgb(r, g, b) => rgb_to_ansi256(r, g, b) == *n,
-        },
+        Expected::Ansi256(n) => cell.to_index() == *n,
         Expected::Hex(er, eg, eb) | Expected::Rgb(er, eg, eb) => {
-            let (er, eg, eb) = (*er, *eg, *eb);
-            match cell {
-                Color::Default => er == 0 && eg == 0 && eb == 0,
-                Color::Idx(i) => {
-                    let (r, g, b) = ansi256_to_rgb(i);
-                    r == er && g == eg && b == eb
-                }
-                Color::Rgb(r, g, b) => r == er && g == eg && b == eb,
-            }
+            let rgb = match cell {
+                Color::Rgb(r, g, b) => (r, g, b),
+                c => ansi256_to_rgb(c.to_index()),
+            };
+            rgb == (*er, *eg, *eb)
         }
     }
 }
 
 /// Render a cell's color in the same space as the expected value, for messages.
-pub fn describe_cell(cell: Color, expected: &Expected) -> String {
+pub fn describe_cell(cell: Option<Color>, expected: &Expected) -> String {
+    let Some(cell) = cell else {
+        return "default".to_string();
+    };
     match expected {
-        Expected::Ansi256(_) => match cell {
-            Color::Default => "0".to_string(),
-            Color::Idx(i) => i.to_string(),
-            Color::Rgb(r, g, b) => rgb_to_ansi256(r, g, b).to_string(),
-        },
-        _ => match cell {
-            Color::Default => "#000000".to_string(),
-            Color::Idx(i) => {
-                let (r, g, b) = ansi256_to_rgb(i);
-                format!("#{r:02x}{g:02x}{b:02x}")
-            }
-            Color::Rgb(r, g, b) => format!("#{r:02x}{g:02x}{b:02x}"),
-        },
+        Expected::Ansi256(_) => cell.to_index().to_string(),
+        _ => {
+            let (r, g, b) = match cell {
+                Color::Rgb(r, g, b) => (r, g, b),
+                c => ansi256_to_rgb(c.to_index()),
+            };
+            format!("#{r:02x}{g:02x}{b:02x}")
+        }
     }
 }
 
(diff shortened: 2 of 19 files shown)


src/monitor.rs
```diff
@@ -152,76 +145,66 @@ fn border_line(out: &mut String, left: char, right: char, title: &str, inner_w:
...
+            let sub = match u.style {
+                UnderlineStyle::Single => 1,
+                UnderlineStyle::Double => 2,
+                UnderlineStyle::Curly => 3,
+                UnderlineStyle::Dotted => 4,
+                UnderlineStyle::Dashed => 5,
+            };
+            s.push_str(&format!(";4:{sub}"));
+            match u.color {
+                Some(Color::Rgb(r, g, b)) => s.push_str(&format!(";58;2::{r}:{g}:{b}")),
```
cpendery: I'm pretty sure this is going to break the SGR handling by mixing the ; & : operators

emitted:   \x1b[0;4:3;58;2::1:2:3;38;5;1m
parsed as: fg=None  underline.color=Rgb(38, 5, 1) , eating the 38;5;1 foreground

src/terminal/alacritty.rs
```diff
@@ -0,0 +1,220 @@
...
+}
+
+fn cell_from_alac(c: &alacritty_terminal::term::cell::Cell) -> EmuCell {
+    let flags = c.flags;
+    let spacer = flags.contains(AlacFlags::WIDE_CHAR_SPACER)
+        || flags.contains(AlacFlags::LEADING_WIDE_CHAR_SPACER);
+    let ch = if spacer {
+        CompactString::const_new(CONTINUATION)
+    } else {
+        let mut s = CompactString::from(c.c.to_string());
```
cpendery: this will result in extra heap allocations, use c.c.to_compact_string() instead

src/terminal/alacritty.rs
```diff
@@ -0,0 +1,220 @@
...
+    Some(Underline {
+        style,
+        color: c.underline_color().and_then(color_from_alac),
+    })
+}
+
+fn cell_from_alac(c: &alacritty_terminal::term::cell::Cell) -> EmuCell {
+    let flags = c.flags;
+    let spacer = flags.contains(AlacFlags::WIDE_CHAR_SPACER)
+        || flags.contains(AlacFlags::LEADING_WIDE_CHAR_SPACER);
```
cpendery: I'm pretty sure this is going to mislabel continuations as this comes before a wide character, 

for something like  "abcd你", it'll move the wide character to line 2, but not fill in the empty space resulting in a short row, in this case, a row of 4 "abcd" instead of "abcd "

bindings/js/src/types.ts
```diff
@@ -24,13 +31,22 @@ export interface Size {
...
+  dim: boolean;
   italic: boolean;
-  underline: boolean;
   inverse: boolean;
+  invisible: boolean;
+  strike: boolean;
+  /** Always `false` from the alacritty backend, which cannot report blink. */
+  blink: boolean;
+  underline: boolean;
+  underline_style: UnderlineStyle | null;
```
cpendery: I don't love the idea of using null for unset, having an enum with a "none" / "default" value makes more sense to me where it's always going to be required

priority work [xtermjs](https://github.com/xtermjs/xterm.js/blob/904ae935269eef5ec6a1415b64463c3d02eff1eb/demo/client/components/window/cellInspectorWindow.ts#L11)

src/assert/color.rs
```diff
@@ -53,43 +53,39 @@ fn parse_hex(hex: &str) -> anyhow::Result<(u8, u8, u8)> {
 }
 
 /// Does a cell's resolved color match the expected color?
-pub fn matches(cell: Color, expected: &Expected) -> bool {
+///
+/// A cell using the terminal default never matches: the default is whatever
+/// the viewer's theme picks, so no concrete value can be claimed for it.
```
cpendery: nit: probably need to cleanup some comments, same in the existing code

src/terminal/alacritty.rs
cpendery: lets add some tests specifically around line wraps / snapshots to make sure we aren't going to break this as we introduce more backends / have any regressions

cpendery (commented):
some small comments on specific code, also generally if `expect --fg 0` isn't working for the default color for the cell, we should enable the user to do `expect --fg default` since default is supported for snapshots & error messages

src/monitor.rs
aymanbagabas: Confirmed, and worse than a cosmetic problem: `;58` ends the parameter, so the terminal read it as a bare "set underline color" and ate the parameters that followed. Round-tripping the old output through the emulator gave `fg=None, underline.color=Rgb(38, 5, 1)` — the underline swallowed the `38;5;1` foreground.

Fixed in 6cbeb08 by emitting the colon form, `58:2::{r}:{g}:{b}` and `58:5:{n}`, consistent with the `4:{sub}` already there. Both pure forms parse correctly; only the mix was broken.

The real gap was that `sgr` output only ever went to a real terminal, where a malformed escape silently reassigns parameters instead of failing, so there was nothing to fail. Added `sgr_survives_a_round_trip_through_the_emulator`, which feeds the output back through the emulator and compares the style that comes out.

src/terminal/alacritty.rs
aymanbagabas: Done in 322519f. Embarrassing, since removing that allocation was the whole point of moving to `CompactString`. Now `c.c.to_compact_string()`.

src/terminal/alacritty.rs
aymanbagabas: Good catch, this was a real bug. The two flags are opposites: `WIDE_CHAR_SPACER` is the second column of a wide char and renders as nothing because the glyph already covered it, while `LEADING_WIDE_CHAR_SPACER` is filler in the last column when a wide char did not fit and wrapped, so it owns its column and has to render as a blank.

Folding them together dropped a column. On a 5-wide grid, `abcd你` reported row 0 as `"abcd"` — four columns for a five-column row, and everything downstream that counts columns saw it short.

Fixed in 322519f; only `WIDE_CHAR_SPACER` maps to `CONTINUATION` now.

bindings/js/src/types.ts
aymanbagabas: Agreed, and it applies to both fields — `fg`/`bg` already spell this idea as `"default"`, so the underline pair was the odd one out. Fixed in f521c186:

- `underline_style` is always present, `"none" | "single" | "double" | "curly" | "dotted" | "dashed"`.
- `underline_color` uses the same `"default"` sentinel as `fg`/`bg` for "follows the text color".

Neither can be null now, so a client switches on the string and compares the color exactly the way it does for `fg`. `underline: bool` stays as a shorthand for `style !== "none"`. Both bindings updated, and the wire name now lives on `UnderlineStyle` so the next backend reuses it rather than re-deriving it in the daemon.

src/assert/color.rs
aymanbagabas: Cleaned up in eff190fa. The `matches` comment had gone stale in the process — it claimed a default cell "never matches", which stopped being true once `default` became an expected value. Rewrote it to say what actually holds: a default cell matches only `default`, never a concrete value, and why.

src/terminal/alacritty.rs
aymanbagabas: Added in 322519f: `conformance_wide_char_wraps_at_the_line_edge` pins the vacated column and the wrapped row, and `conformance_wrap_snapshot` pins a whole multi-row grid. Both live in the shared conformance macro, so ghostty and xterm.js have to draw the same distinction to pass.

bindings/js/src/types.ts
aymanbagabas: Agreed, and it applies to both fields — `fg`/`bg` already spell this idea as `"default"`, so the underline pair was the odd one out. Fixed in f521c186:

- `underline_style` is always present, `"none" | "single" | "double" | "curly" | "dotted" | "dashed"`.
- `underline_color` uses the same `"default"` sentinel as `fg`/`bg` for "follows the text color".

Neither can be null now, so a client switches on the string and compares the color exactly the way it does for `fg`. `underline: bool` stays as a shorthand for `style !== "none"`. Both bindings updated, and the wire name now lives on `UnderlineStyle` so the next backend reuses it rather than re-deriving it in the daemon.

aymanbagabas:
All six addressed, pushed as four commits.

**`expect --fg default`** — you were right that this was the missing half. Snapshots and failure messages had been calling it `default` all along, but there was no way to assert on it, so a default cell matched nothing. `Expected::Default` now accepts the same word (eff190fa): `--fg default` passes only when the cell set no foreground, and a cell that did set one fails against it, so it reads both ways. Refusing `--fg 0` for a default cell stays refused — which color the theme paints it is not something the grid knows.

**Two of your comments were real bugs, both confirmed before fixing:**

- The SGR one was worse than malformed. Round-tripping the old output through the emulator gave `fg=None, underline.color=Rgb(38, 5, 1)` — the `;` ended the parameter and the underline color ate the `38;5;1` foreground. It went unnoticed because `sgr` only ever wrote to a real terminal, where a bad escape silently reassigns parameters rather than failing.
- The spacer one dropped a column: `abcd你` on a 5-wide grid reported row 0 as `"abcd"`.

Both now have the check that was missing — an SGR round-trip test, and wrap tests in the shared conformance macro so ghostty and xterm.js have to get this right too.

| | |
|---|---|
| `LEADING_WIDE_CHAR_SPACER` + wrap tests + `to_compact_string` | 322519f |
| SGR `58` colon form + round-trip test | 6cbeb08 |
| Non-nullable `underline_style` / `underline_color` | f521c18 |
| `expect --fg default` + comment cleanup | eff190f |

85 tests pass, clippy and fmt clean, both bindings build.

src/terminal/alacritty.rs
aymanbagabas: Follow-up — you asked for line wraps **and snapshots**, and I had only half-delivered on the snapshot part.

What I first added tested the two sides separately: `conformance_wide_char_wraps_at_the_line_edge` pinned the grid, and a unit test in `assert/snapshot.rs` pinned `serialize` against hand-built rows. Both passed. But nothing ran the serializer over a grid a backend actually produced, and that composition is exactly what broke in the first place: the grid was right, `serialize` was right on its own inputs, and the frame still came out a column wide because the two disagreed on what a continuation renders as.

`conformance_snapshot_frames_a_wrapped_wide_char` (d784a7a) closes that. It is in the conformance macro, so every backend has to pass it. Verified it fails on the real bug rather than just passing today:

```
left:  "╭─────╮\n│abcd │\n│你    │\n│     │\n╰─────╯"   ← continuation rendered as a space
right: "╭─────╮\n│abcd │\n│你   │\n│     │\n╰─────╯"
```

That turned up a third instance of the same bug, by the way — `snapshot.rs` was still mapping a continuation to `" "`, so every saved snapshot containing a wide char was written misaligned and `--update` baked it in as the baseline. Fixed in 8f84dc3.

Also renamed `conformance_wrap_snapshot` to `conformance_wrap_grid_over_several_rows`; it dumps the grid and never touched the snapshot code its name implied.

src/assert/color.rs
```diff
@@ -12,6 +14,9 @@ pub enum Expected {
 impl Expected {
     pub fn parse(s: &str) -> anyhow::Result<Self> {
         let s = s.trim();
+        if s.eq_ignore_ascii_case("default") {
```
cpendery: nit: maybe we can define this as a constant so we don't keep magic stringing everywhere?

src/assert/snapshot.rs
```diff
@@ -43,43 +43,27 @@ fn shift(prev: &EmuCell, cur: &EmuCell) -> Map<String, Value> {
...
+        (Attrs::BLINK, "blink"),
+    ] {
+        if prev.has(attr) != cur.has(attr) {
+            m.insert(key.into(), json!(cur.has(attr)));
+        }
     }
-    if prev.strike != cur.strike {
-        m.insert("strike".into(), json!(cur.strike));
+    if prev.underline.is_underlined() != cur.underline.is_underlined() {
+        m.insert("underline".into(), json!(cur.underline.is_underlined()));
```
cpendery: we could update this to show the underline style instead of just is underlined since we added support for that

cpendery (approved):

cpendery (approved):

PR #79, opened 2026-07-31
feat: add new backend, rio-vt

raphamorim:
This swaps the VT emulation backend from alacritty_terminal to [rio-vt](https://crates.io/crates/rio-vt), the terminal core we extracted from [Rio](https://github.com/raphamorim/rio). The PTY layer (portable-pty) is untouched, so this is purely an engine swap behind `Emu`.

What changed:

- `emu.rs` now wraps `rio_vt::crosswords::Crosswords`. Cell colors/attributes are resolved from rio-vt's packed cell + per-grid style table, mapped to the same `EmuCell`/`Color` surface the rest of the code consumes.
- `CommandTracker` in `integration.rs` keeps its exact logic; it just implements rio-vt's `Perform` trait instead of vte's (same state machine lineage, same callback shape, so the diff there is one import line).
- Terminal replies (DA/DSR/CPR) come from rio-vt as events and flow through the same pending-writes path as before.

Why: rio-vt is actively maintained, it's the engine we run in production in Rio, and it brings correct unicode width/grapheme handling plus shell-integration parsing (OSC 7/133) natively should shell-use want to lean on more of it later. It also drops the winit-adjacent transitive deps that alacritty_terminal pulls in. Throughput is on par with or ahead of alacritty_terminal on parsing-heavy workloads; benchmarks against alacritty_terminal and vt100 live at https://github.com/raphamorim/rio-vt-benchmark.

Verified: all 48 tests pass, clippy clean. I also drove the real daemon on macOS: session open, `submit`/`wait command`, exit-code + cwd tracking (OSC 133/7), bold/color/underline cell attributes via `cells`, and wide characters (CJK with correct spacer cells). Screenshot/SVG rendering works.

One pre-existing quirk I noticed while testing (not from this change; reproduces on main): with zsh the captured command can double the first character ("pprintf ..."), which seems to come from prompt repaint echo. Happy to dig at that separately.


cpendery:
Hey @raphamorim, thank you so much for the PR, let's hold on this until #78 merges. We are working on adding support for multiple backends, and it would be great to have rio as an option

raphamorim:
Sounds good!! Thanks for the reply

cpendery:
@raphamorim, now that we also everything setup to support ghostty, we should be good to add rio-vt as an alternative backend. Would you like to continue with the changes or I can pick this up and add them to this branch

raphamorim:
If you could continue I would highly appreciate, otherwise I can pick up in a near future 🙏

PR #83, opened 2026-08-04
feat(terminal): add xterm.js as a selectable backend

aymanbagabas:
tui-test can already run a session on alacritty, ghostty, or rio. This adds the emulator behind VS Code's terminal as a fourth, so a suite can be run against what a large share of users are actually looking at. A test that passes on one emulator and fails on another is telling you something real about the program under test.

```console
$ tui-test open --backend xtermjs
$ tui-test run --backend xtermjs -- vim notes.md
```

Selectable everywhere the other backends are: the CLI's `--backend`, and both the Python and Node bindings.

## No Node at runtime

`@xterm/headless` and the unicode11 addon are vendored and evaluated into a QuickJS context per session, so the backend depends on nothing installed on the machine. `shim.js` is our own code; the bundles are dropped in unchanged at pinned versions.

They are committed rather than fetched, so a clone builds without a network and there is no build script. `.github/scripts/vendor-xtermjs.sh --latest` bumps `pinned.json` to the newest releases and re-fetches in one command, and `.github/workflows/xtermjs-update.yml` runs that same script weekly and opens a pull request when either package has a newer release.

The grid crosses that boundary packed rather than a cell at a time. Reading an 80x30 screen through the per-cell getters is 2,400 calls with ten property reads each, which costs milliseconds per poll; the shim flattens a row span into one string and one integer array, so a whole screen crosses as two values and this side decodes rather than traverses.

## What the shim has to supply

Five things the `Emulator` contract asks for are missing from, or wrong in, the headless bundle's public surface. The shim supplies them rather than the Rust side pretending they are absent:

| Contract | How |
| --- | --- |
| Window title | `onTitleChange`, with `windowOptions.pushTitle`/`popTitle` enabling the `CSI 22/23 t` stack the bundle implements but leaves off by default |
| Cursor visibility | `coreService.isCursorHidden` |
| Cursor shape | `coreService.decPrivateModes.cursorStyle`, which is absent until `DECSCUSR` sets one and so reads as a block until then |
| `OSC 4/10/11/12` set, query, and reset | `parser.registerOscHandler`, answering out of the same reply queue the terminal's own replies use, so answers keep the order they were asked in |
| `SGR 59` | xterm.js stores the reset as an explicit white, indistinguishable from a real `58;2;255;255;255` once written; the shim clears it where the parser sets it, while the two are still apart |

A colour reply has to echo the terminator its query used, and an OSC handler is handed its payload but not that terminator. The shim scans the incoming bytes for it, carrying state between calls because a PTY read splits wherever it likes — including between the two bytes of an `ST` — and keying what it finds by OSC code so a title arriving between two queries cannot put the wrong terminator on a reply.

Verified against a real session rather than only in unit tests. The same probe run on both backends, querying the background, setting it, resetting it, and querying once more with each terminator:

```
alacritty:  <E>]11;rgb:0000/0000/0000<BEL> | <E>]11;rgb:6565/4343/2121<BEL> | <E>]11;rgb:0000/0000/0000<BEL> | ST:<E>]11;rgb:0000/0000/0000<E>\
xtermjs:    <E>]11;rgb:0000/0000/0000<BEL> | <E>]11;rgb:6565/4343/2121<BEL> | <E>]11;rgb:0000/0000/0000<BEL> | ST:<E>]11;rgb:0000/0000/0000<E>\
```

Byte-identical, terminator included.

## Conformance

The backend passes the conformance suite, which is what makes swapping emulators safe, with **one declared exception**: xterm.js records a cell's underline colour only when that cell also has an underline style, so `SGR 58` on its own — or a colour that outlives an `SGR 24` — is not readable back off the cell.

Nothing renders differently, since a cell with no underline draws no underline colour either way. What is lost is the colour surviving in the cell vocabulary. I confirmed this against the bundle rather than inferring it: the cell reports `isAttributeDefault()`, and the colour is gone from the line's extended attributes while remaining in the current SGR state, so it is xterm.js's per-cell storage rather than anything this mapping does.

The suite grew a `Divergences` declaration for it, so the exception is a visible claim a reviewer can check rather than a quietly failing case or a backend that skips conformance altogether. It is deliberately narrow: a field earns a place there only when the limitation is in the emulator itself.

## Also here

`Backend::ALL` becomes a slice, so each backend adds one line rather than doubling the number of cfg-gated definitions, and the list of names in a parse error is derived from it instead of repeated in prose that can drift out of step.

## Notes for review

- **The backend is compiled in by default** for the CLI and both bindings, matching how ghostty is wired, so `--backend xtermjs` works out of the box. Measured cost: 229.6 KiB of embedded JavaScript, and a release binary going 9,208,800 -> 10,506,064 bytes, so about 1.30 MB once QuickJS is compiled in. Happy to put it behind an off-by-default feature instead if the size matters more than the availability.
- The `xtermjs` crate feature itself is off by default, so `cargo build -p tui-test-rs` is unaffected.
- Conformance for the optional backends does not run in CI today, since `cargo test --workspace` uses default features while only clippy passes `--all-features`. That predates this PR and applies to ghostty and rio equally, but it does mean these 53 cases are green locally rather than on a runner. Worth a follow-up.
- Unicode 11 is pinned deliberately, not incidentally: it is the only version that agrees with alacritty on every emoji width case measured. The table is in `assets/xtermjs/README.md`.

## Review

Four reviewers went over this (shim correctness, Rust/FFI, architecture, supply chain). Three real divergences came back, all now fixed in the second commit and all reproduced before being touched:

- **A split `ST` was answered with `BEL`.** The scan that recovers a query's terminator waited for the `\` of an `ST`, but xterm.js ends an OSC on the `ESC` before it, so a PTY read splitting between those two bytes answered the wrong terminator. Recording it on the `ESC` keeps the two in step.
- **The terminator queue grew without bound.** It recorded every OSC sequence while only colour ones claim what it records, so a shell that retitles the window on each prompt leaked an entry per prompt. Both the shim and supply-chain reviewers found this independently.
- **`OSC 4;1x` wrote to slot 1.** `parseInt` took the leading digit of an index that was not a number.

The first two are contract behaviour rather than quirks of this backend, so they went into the conformance suite: both fail on xterm.js without the fix and pass on alacritty, which is what makes them worth asserting of every backend.

Also fixed: a `usize` -> `u32` scrollback cast that wrapped a deep request into a shallow one, and the vendored licence, which reproduced only one of the two packages' notices.

Verified clean by review, worth recording:

- **Bundle provenance is byte-for-byte.** Both vendored files match what npm publishes for `@xterm/headless@6.0.0` and `@xterm/addon-unicode11@0.9.0`, reproducible with `.github/scripts/vendor-xtermjs.sh`.
- **No path from terminal bytes to evaluated code.** Only the three static assets are ever evaluated; PTY bytes reach JS as a byte array.

Considered and **not** done, happy to be overruled:

- **A QuickJS memory cap.** The one unbounded growth path was the queue above, and it is fixed; what remains is bounded by scrollback. A cap would need a number I have no evidence for, and one set too low breaks a legitimately deep session.
- **Bell tracking.** xterm.js exposes `onBell`, but `build_with_bells` falls through to `build` exactly as ghostty's does. Worth doing for both backends together rather than making them differ.

## Review round two

Ten threads from @cpendery, all addressed and resolved. Two of them landed the other way from the review; both are called out below with the reasoning, and I am happy to reopen either.

**Two were real bugs**, both reproduced before being touched:

- **`SGR 59` swallowed a genuine white underline.** xterm.js stores the reset as an explicit RGB #ffffff, and every public getter then reports exactly what a real `58;2;255;255;255` reports, so reading the cell could only guess — and the old heuristic guessed "reset", meaning no cell could ever have a white underline. The shim now clears it where the parser sets it, while the two are still distinguishable. Verified the new conformance case fails on xterm.js with the old heuristic restored.
- **JS exceptions were suppressed.** `process` and `resize` discarded their `Result` entirely, so a throw froze the grid and every later read answered from it as fact. Faults are now recorded and reported by the next operation, through a new `Emulator::fault` checked in `Engine::with_session` — before the operation runs, since a wrong answer off a stale grid is worse than a failure. This is not hypothetical: the shim reads several private xterm.js fields, so a bundle upgrade moving one is exactly how it fires.

**Two went against the review**, so the reasoning is here rather than buried in a thread:

- **Vendoring.** I tried de-vendoring in 0fc1fbb and reverted it in 8ff376c: a build script put npm and a network on the path of anyone building from a clone, which is a heavier cost than keeping 264 KB out of git. The bundles stay committed, and `.github/scripts/vendor-xtermjs.sh --latest` bumps and re-fetches in one command.
- **How a stale pin is noticed.** A weekly pull request rather than a release-time check. A release gate fires on someone else's schedule — an upstream publish turns a ready tag into a failed release — and a version-string comparison cannot see the failure that actually matters: the shim reads several private xterm.js fields, so a release moving one is the real risk. The weekly job runs the conformance suite against the new bundles before opening the PR, which nothing else does, since CI builds with default features. It opens one only on a real new release, decided from the vendored bytes rather than `pinned.json` alone.

  I also tried and reverted a release-time provenance check (bcb3c650 → 50b5d7e): it re-asserted something already settled when the files were written, for a mismatch that could not reach the artifact, and put a network dependency on the release path. Nothing xterm.js-specific runs at release time now.

**The rest:**

- Added conformance cases for xterm.js's special cases: the underline reset above, and the `isAttributeDefault()` fast path where one call stands in for nineteen getters — that shortcut previously had a comment claiming it was verified with nothing checking it.
- Trimmed `assets/xtermjs/README.md` from 114 lines to 74, and moved the deferred-addons section to #175.
- Dropped the `Backend::ALL` comment.
- Answered the `raster.rs` question (recent clippy lints, isolated in 9a68c1c) and the OSC 10-12 one (fixed earlier in 8f6520b, with two conformance cases).

Every new conformance case is a case, not a snapshot of current behaviour: each was verified to fail on xterm.js without its fix and pass on alacritty, ghostty, and rio.

## Testing

53 conformance cases against the xterm.js backend, plus the full workspace suite (514 tests), `cargo clippy --workspace --all-targets --all-features -- -D warnings`, the no-default-features clippy pass, fmt, and the Python and Node binding suites. `cargo publish --dry-run` packages 94 files and compiles from the tarball.

Also verified that a clone builds with no network and no Node: `cargo build -p tui-test-rs --features xtermjs` succeeds with npm off `PATH`, which is the property the vendored bundles exist to preserve.

Driven end to end through a real PTY as well: title, cursor, colour assertions, and SVG screenshots all agree with alacritty.


crates/shell-use/assets/xterm/README.md
```diff
@@ -0,0 +1,40 @@
+# Vendored xterm.js assets
```
cpendery: I'm wondering if we can just use the shim + esbuild & pinned versions for xtermjs's npm module. I'd like to avoid vendoring where we can

cpendery (commented):
Overall looks good, I think we can put alacritty as the default & include the other backends as features. For the client libraries, I think it makes sense to include xtermjs & alacritty by default

crates/shell-use/assets/xterm/README.md
aymanbagabas: Agreed on the instinct, and I dug into what it would actually buy us here. Two things make this case narrower than it looks:

**There's nothing to bundle.** `@xterm/headless` publishes `lib-headless/xterm-headless.js` already bundled and minified by its own webpack build. What's checked in is that file, copied out of the tarball unchanged — so esbuild would be re-bundling an artifact that's already a bundle. `shim.js` is our code and is already separate.

**The versions are already pinned**, just pinned in git rather than in a lockfile. Same two packages, same exact bytes:

```
xterm-headless.js   @xterm/headless@6.0.0          byte-for-byte
addon-unicode11.js  @xterm/addon-unicode11@0.9.0   byte-for-byte
```

The real question is only whether the bytes live in the repo or get fetched during the build, and fetching costs more than it first appears: `cargo build --features xtermjs` would need network and a Node toolchain, which breaks offline and air-gapped builds and makes publishing to crates.io awkward. That's a heavier build requirement than Ghostty's Zig, which we already call out in the README as the reason Windows ARM64 artifacts aren't published. It'd also mean the one backend that currently needs no external toolchain would start needing two.

So I've gone at the underlying worry instead — that a vendored blob can silently stop matching what it claims to be. There's now a CI job that re-fetches both packages at their pinned versions and fails if what's checked in isn't byte-identical, and it only runs on commits that touch `assets/xterm/`, so it costs nothing elsewhere. The SHA-256s and the exact commands to reproduce the files are in the README next to them. A hash committed alongside the bytes it describes proves nothing on its own; re-fetching from npm is what actually closes that loop.

Worth noting we already vendor 6.6 MB of fonts under the same `crates/tui-test/assets/` with adjacent licenses. xterm.js is 264 KB, about 4% of that, and follows the same shape.

Happy to switch to a build-time fetch if you'd still prefer it — just wanted the toolchain cost on the table first, since it lands on everyone building the crate.

aymanbagabas:
> I think we can put alacritty as the default & include the other backends as features.

That's already how the crate is set up — `tui-test-rs` has `default = []`, so a plain `cargo add tui-test-rs` gets Alacritty only, and both `ghostty` and `xtermjs` are opt-in features. Nothing changed there in this PR.

> For the client libraries, I think it makes sense to include xtermjs & alacritty by default

This PR does the additive half: `xtermjs` is now enabled for the CLI and both the Python and Node native packages, so `--backend xtermjs` and `backend="xtermjs"` work out of the box.

The other half — dropping Ghostty from the published bindings — I've deliberately left out, and I think it deserves its own PR. It's a change to what shipped packages contain rather than an addition, it touches a backend this PR doesn't otherwise go near, and it has a release-notes consequence for anyone already passing `backend="ghostty"`. Folding it in here would put two unrelated stories in one diff.

It's also worth doing on its own merits, and the argument is stronger than just build times: Ghostty needs Zig 0.16 to build, and its upstream Zig build is the reason we don't publish Windows ARM64 artifacts at all today. xterm.js needs no external toolchain, so bindings carrying Alacritty + xterm.js could publish everywhere the CLI does. That looks like the real motivation for your suggestion, and it's a good one — I'd just rather land it where it can be reviewed as that change.

Happy to open it as a follow-up right after this, or to pull it in here if you'd rather have them together.

cpendery:
> > I think we can put alacritty as the default & include the other backends as features.
> 
> That's already how the crate is set up — `tui-test-rs` has `default = []`, so a plain `cargo add tui-test-rs` gets Alacritty only, and both `ghostty` and `xtermjs` are opt-in features. Nothing changed there in this PR.
> 
> > For the client libraries, I think it makes sense to include xtermjs & alacritty by default
> 
> This PR does the additive half: `xtermjs` is now enabled for the CLI and both the Python and Node native packages, so `--backend xtermjs` and `backend="xtermjs"` work out of the box.
> 
> The other half — dropping Ghostty from the published bindings — I've deliberately left out, and I think it deserves its own PR. It's a change to what shipped packages contain rather than an addition, it touches a backend this PR doesn't otherwise go near, and it has a release-notes consequence for anyone already passing `backend="ghostty"`. Folding it in here would put two unrelated stories in one diff.
> 
> It's also worth doing on its own merits, and the argument is stronger than just build times: Ghostty needs Zig 0.16 to build, and its upstream Zig build is the reason we don't publish Windows ARM64 artifacts at all today. xterm.js needs no external toolchain, so bindings carrying Alacritty + xterm.js could publish everywhere the CLI does. That looks like the real motivation for your suggestion, and it's a good one — I'd just rather land it where it can be reviewed as that change.
> 
> Happy to open it as a follow-up right after this, or to pull it in here if you'd rather have them together.

based on our offline sync, I think we want to include all backends in the bindings.

crates/shell-use/assets/xterm/README.md
cpendery: I'd still like to avoid vendoring, I think its a little different from the fonts since we can just use npm to fetch them. 

I'm reworking the readme, and we can add a development section that covers the requirements of both zig & node for building. Let's still bundle them into the published crate, just not checked into the repo

aymanbagabas: We need to bundle xtermjs into the published crate whether we're vendoring its assets in the repository or fetching them during `cargo build`. Vendoring them in the repository is safer because you don't depend on npmjs or whatever registry and network to fetch the files during build, and the files are already in the repository here so during build the compiler can bundle them right away as static assets.

So really our other option is fetching them during `cargo build` which is more error prone because now you have network build dependency, and you need to handle xtermjs version tracking somehow perhaps using `build.rs`.

If you still insist on the network approach, I'm happy to make the changes 🙂

.github/workflows/xtermjs-update.yml
```diff
@@ -0,0 +1,138 @@
+name: Update xterm.js
```
cpendery, line 1: I'd rather we just bump it manually when needed instead of having the noise of weekly pr's. similar to how node-pty / portable-pty has a vendored version of conpty they bump when needed. We should instead add a step to the release pipeline in the verify to fail unless the version if the most recent latest release

crates/tui-test/assets/xtermjs/README.md
```diff
@@ -0,0 +1,114 @@
...
+xterm.js records a cell's underline colour only when that cell also has an
+underline style: `SGR 58` on its own, or a colour that outlives an `SGR 24`, is
+not readable back off the cell. Nothing renders differently, since a cell with
+no underline draws no underline colour either way. The backend declares this
+where it opts into the conformance suite, so the exception is visible rather
+than silent.
+
+## Addons we do not use yet
+
+Two are worth revisiting, neither of them yet:
```
cpendery: make an issue to track this, not the readme

crates/tui-test/assets/xtermjs/README.md
```diff
@@ -0,0 +1,114 @@
+# Vendored xterm.js assets
+
+Compiled into the `tui-test` binary by `crates/tui-test/src/terminal/xtermjs.rs`
```
cpendery, line 3: can we keep this fairly brief? it's a lot of info here. I'd like to just keep the explanation of the pinned version very short & the divergence section also shortened up a little

crates/tui-test/src/terminal/backend.rs
```diff
@@ -18,19 +20,31 @@ pub enum Backend {
...
 }
 
 impl Backend {
-    #[cfg(not(feature = "ghostty"))]
-    pub const ALL: [Self; 1] = [Self::Alacritty];
-    #[cfg(feature = "ghostty")]
-    pub const ALL: [Self; 2] = [Self::Alacritty, Self::Ghostty];
+    /// Every backend this build can construct.
+    ///
+    /// A slice rather than a fixed-size array so that each optional backend
```
cpendery: let's drop this comment, not super useful, slice seems better though

crates/tui-test/src/render/raster.rs
```diff
@@ -316,7 +316,7 @@ impl FrameRenderer for GridRenderer {
         }
 
         let mut pixels = self.pixmap.data().to_vec();
-        for pixel in pixels.chunks_exact_mut(4) {
+        for pixel in pixels.as_chunks_mut::<4>().0 {
```
cpendery, line 319: what's this change for?

crates/tui-test/assets/xtermjs/shim.js
```diff
@@ -0,0 +1,381 @@
...
+          answer(prefix, resolved(slot), bel);
+        } else {
+          var v = parseColor(parts[i]);
+          if (v !== null) { overrides[slot] = v; }
+        }
+      }
+      return true;
+    });
+  }
+  dynamic(10, FG);
```
cpendery: I'm  pretty sure OSC 10-12 support multi-color emissions like `ESC ] 10 ; #010203 ; #040506 BEL`  which can change foreground & background color together. This parser `parseColor` won't handle that properly

crates/tui-test/assets/xtermjs/shim.js
```diff
@@ -0,0 +1,381 @@
...
+
+          // SGR 59 (reset underline color) does not clear the record: it
+          // stores a sentinel that reads back through the public getters as
+          // RGB #ffffff, so an ordinary reset produced a white underline where
+          // there should be none. The sentinel is indistinguishable from a
+          // real `58;2;255;255;255` at this layer -- both report RGB with
+          // value 0xffffff -- so one of the two has to be wrong. Resetting is
+          // overwhelmingly the more common of the two, and getting it wrong
+          // paints a color the terminal never asked for, so it wins.
+          if (ulMode === 2 && ulColor === 0xffffff) { ulMode = 0; }
```
cpendery: doesn't this cause issues when a real white underline is emitted?

crates/tui-test/assets/xtermjs/shim.js
```diff
@@ -0,0 +1,381 @@
...
+          // record that it drops whenever the underline style is NONE, and
+          // both underline-color getters then fall back to reporting the
+          // foreground. Left alone that shows up as every colored cell
+          // claiming an underline color it was never given. Collapsing the
+          // case where the two are identical back to "unset" is exactly the
+          // vocabulary's own spelling for it: `underline_color: None` already
+          // means the underline takes the foreground. A cell that really did
+          // set SGR 58 to its own foreground color lands here too, and draws
+          // the same either way.
+          if (ulColor === fg && ulMode === fgMode) { ulMode = 0; }
```
cpendery, line 383: can we add some conformance cases to ensure as much as possible that xterm's special cases are properly behaving?

crates/tui-test/src/terminal/xtermjs.rs
```diff
@@ -0,0 +1,421 @@
...
+        self.cols = self.call::<i32>("cols").clamp(0, u16::MAX as i32) as u16;
+        self.rows = self.call::<i32>("rows").clamp(0, u16::MAX as i32) as u16;
+    }
+
+    /// Call a zero-argument method on the shim's emulator object.
+    ///
+    /// No `this` is threaded through: every method the shim returns is a
+    /// closure over its own `term`, so the receiver is unused, and rquickjs
+    /// would otherwise pass a `This` wrapper as the first positional argument.
+    fn call<R>(&self, method: &str) -> R
```
cpendery, line 218: doesn't this swallow all errors? and don't process and resize ignore exceptions entirely? I think we need to find a better way to surface the failures instead of just suppressing them

cpendery (commented):
left some comments on some fixes, overall work looks really good. 

nit: can we make sure the ci is actually running the xtermjs conformance tests? cargo test --workspace uses the default features so it'll skip the xterm tests even if clippy compiles them

crates/tui-test/src/render/raster.rs
aymanbagabas: lint in recent rust clippy version

crates/tui-test/assets/xtermjs/shim.js
aymanbagabas: Do they? I know OSC 10 -> Fg, OSC 11 -> Bg, and OSC 12 -> cursor. Here's the specs https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands

aymanbagabas: You're right, and my earlier reply was wrong — sorry. The same page I linked says it a little further up, in the preamble to the `Ps = 1 0` list rather than on the entries themselves, which is how I missed it:

> At least one parameter is expected for `Pt`. **Each successive parameter changes the next color in the list. The value of `Ps` tells the starting point in the list.**

and for queries:

> Because more than one pair of color number and specification can be given in one control sequence, xterm can make more than one reply.

So `Ps` is an index into one list of dynamic colors, not a fixed target. Alacritty already reads it that way, which made this straightforward to confirm — same sequence, both backends:

```
OSC 10;#010203;#040506
  alacritty:  fg=rgb(1,2,3)   bg=rgb(4,5,6)     ← correct
  xtermjs:    fg=rgb(4,5,6)   bg=unset          ← last parameter won
```

Queries were wrong in the same shape — `OSC 10;?;?` answered `OSC 10` twice with the foreground, instead of `OSC 10` then `OSC 11`.

Fixed in 8f6520b. Each parameter now advances to the next color, and each `?` is answered under the code of the color it actually asked about. The list runs past the cursor into pointer, Tektronix, and highlight colors we have no notion of; those positions are stepped over rather than folded onto the last slot, so a long sequence sets what it names and nothing else.

Both went in as conformance cases rather than xterm.js-only tests, since this is contract behaviour: they fail on xterm.js without the change and pass on alacritty.

```
xtermjs  conformance_one_dynamic_color_sequence_sets_successive_colors ... FAILED
  left: Rgb { r: 4, g: 5, b: 6 }   right: Rgb { r: 1, g: 2, b: 3 }
xtermjs  conformance_each_dynamic_color_query_is_answered_separately ... FAILED
  left:  "\e]10;rgb:0404/0505/0606\a\e]10;rgb:0404/0505/0606\a"
  right: "\e]10;rgb:0101/0202/0303\a\e]11;rgb:0404/0505/0606\a"
```

Good catch — this is exactly the kind of thing the conformance suite should have been covering and wasn't.

crates/shell-use/assets/xterm/README.md
aymanbagabas: Done — de-vendored in 0fc1fbb. You were right that npm makes this different from the fonts.

`crates/tui-test/build.rs` reads `pinned.json`, fetches both packages from npm, and caches them in `OUT_DIR` keyed by version. The bundles are gone from the repo and gitignored.

The published crate still carries them, so the network dependency lands only on people building from a checkout:

- `build.rs` prefers a copy sitting next to it and only fetches when there isn't one. In a published crate there always is.
- Cargo discovers package files through git, which would have dropped gitignored files from the tarball, so `crates/tui-test/Cargo.toml` now lists `include` explicitly. Verified the packaged file set is otherwise byte-identical to before — same 94 files plus `build.rs`.
- `.github/scripts/vendor-xtermjs.sh` puts them on disk before `cargo publish`, and both the verify and publish jobs run it.

Checked both directions rather than assuming:

```
bundles absent  → cargo build --features xtermjs   → fetched, sha256 17a90b65… (identical to what was vendored)
bundles present → PATH without npm                 → builds, no fetch attempted
cargo publish --dry-run                            → Packaged 95 files, compiles from the tarball
```

README has the development section you mentioned: a table of what each optional backend needs to build (Zig for `ghostty`, Node for `xtermjs`, nothing for `rio`), and a note that consuming a release needs neither.

One consequence worth flagging: `cargo publish` now needs `--allow-dirty`, because the files it must package are deliberately gitignored. That's in both release jobs.

.github/workflows/xtermjs-update.yml
aymanbagabas: Done in 0fc1fbb — `.github/workflows/xtermjs-update.yml` is deleted, no more weekly PRs.

The release `verify` job now fails unless both pinned versions are the latest published ones:

```yaml
- name: Verify vendored xterm.js is current
  run: |
    for pkg in @xterm/headless @xterm/addon-unicode11; do
      have="$(jq -r --arg p "$pkg" '.[$p]' pinned.json)"
      want="$(npm view "$pkg" version)"
      [ "$have" = "$want" ] || { echo "$pkg is pinned to $have but $want is the latest release." >&2; status=1; }
    done
```

It reports every stale package rather than stopping at the first, so one run tells you everything to bump.

Bumping by hand is `pinned.json` plus `.github/scripts/vendor-xtermjs.sh`, which is in the README's new development section.

I also dropped `.github/workflows/vendored.yml`. It re-fetched the checked-in bundles and diffed them to catch one being edited or taken from somewhere other than the pinned release — which cannot happen now that they aren't checked in and `build.rs` fetches them itself.

crates/tui-test/assets/xtermjs/README.md
aymanbagabas: Done — filed as #175 and the section is out of the README (0fc1fbb).

The issue keeps both addons and the reason each is deferred: `addon-unicode-graphemes` is still experimental, needs an `atob` the QuickJS host lacks, and disagrees with alacritty on two width cases; `addon-clipboard` would make `OSC 52` answered on one backend and dropped on every other, so the contract and conformance suite want it first.

The README now just links to it.

crates/tui-test/assets/xtermjs/README.md
aymanbagabas: Trimmed in 0fc1fbb — 118 lines down to 81, and a good part of what remains is the width table.

- **Pinned versions**: was a paragraph on `pinned.json` being the single source of truth plus a block of sha256s. Now two sentences, and the hashes are gone — they were only meaningful while the bytes were checked in next to them, and `build.rs` fetches by version now.
- **Known divergence**: cut to the `SGR 58` case and the fact that nothing renders differently.
- **Addons we do not use yet**: moved to #175 per your other comment.
- **OSC terminator paragraph**: dropped the last clause about which byte the terminator is recorded on.

Kept the width table and the "why unicode11" section — that one is load-bearing, since agreeing with alacritty on emoji width is the whole reason for the pin and nothing else asserts it.

crates/tui-test/src/terminal/backend.rs
aymanbagabas: Dropped in 0fc1fbb — slice stays, comment is gone.

```rust
pub const ALL: &'static [Self] = &[
    Self::Alacritty,
    #[cfg(feature = "ghostty")]
    Self::Ghostty,
    #[cfg(feature = "rio")]
    Self::Rio,
    #[cfg(feature = "xtermjs")]
    Self::Xtermjs,
];
```

crates/tui-test/src/render/raster.rs
aymanbagabas: `manual_slice_size_calculation` / the `as_chunks` lints, which recent clippy raises on the constant-size chunk walks. CI runs `cargo clippy --workspace --all-targets --all-features -- -D warnings`, so it was a hard failure rather than a style preference.

It is isolated in 9a68c1c ("fix: adopt as_chunks for the constant-size chunk walks") if you want it separate from the backend work. No behaviour change — `as_chunks` returns the same fixed-size windows the manual arithmetic was computing, with the remainder split off instead of bounds-checked per iteration.

Resolving since it is answered; happy to reopen if you would rather it moved to its own PR.

crates/tui-test/assets/xtermjs/shim.js
aymanbagabas: Fixed in 8f6520b (before the merge), and confirmed still passing on the current head.

Both cases are conformance tests rather than xterm.js-only ones, since this is contract behaviour — they fail on xterm.js without the change and pass on alacritty, ghostty, and rio:

```
conformance_one_dynamic_color_sequence_sets_successive_colors
conformance_each_dynamic_color_query_is_answered_separately
```

Each parameter now advances to the next color in the list, and each `?` is answered under the code of the color it actually asked about. The list runs past the cursor into pointer, Tektronix, and highlight colors we have no notion of; those positions are stepped over rather than folded onto the last slot, so a long sequence sets what it names and nothing else.

crates/tui-test/assets/xtermjs/shim.js
aymanbagabas: Yes it did — good catch. A real white underline was swallowed, and the comment defending it was reasoning from the wrong place.

The two are genuinely identical once written. `SGR 59` does not clear xterm.js's stored color, it writes RGB #ffffff, and every public getter then reports exactly what a real `58;2;255;255;255` reports:

```
                        raw extended.underlineColor   getUnderlineColor()   isUnderlineColorRGB()
58;2;255;255;255        0x3ffffff                     16777215              true
59 (reset)              0x3ffffff                     16777215              true
```

So reading the cell could only ever guess, and the old heuristic guessed "reset" — which is the common case, but it meant no cell could ever have a white underline.

Fixed in 0fc1fbb by correcting it where the parser sets it, while the two are still distinguishable, rather than at read time when they no longer are. The shim wraps `charAttributes` and clears the record on a bare `59`:

```js
if (params.params[i] === 59 && !params.hasSubParams(i)) {
  term._core._inputHandler._curAttrData.extended.underlineColor = 0;
}
```

`0` is the same value the record already holds for a cell that never named an underline color, so this puts the reset back on the path it should have taken. The read-time `0xffffff` heuristic is deleted.

```
                        raw       reported
58;2;255;255;255        0x3ffffff white          ← now survives
59 (reset)              0x0       unset
58;2;255;0;0            0x3ff0000 red
58;5;33                 0x2000021 palette 33
```

There is a conformance case for it, `conformance_resetting_the_underline_color_is_not_white`, covering set → reset → white in one line. It passes on all four backends, and I checked it fails on xterm.js with the old heuristic restored, on exactly the assertion you predicted:

```
assertion `left == right` failed: a white underline is still addressable after a reset
  left: None
 right: Some(Rgb(255, 255, 255))
```

The other special case in that function, folding the underline color back to unset when it equals the foreground, is unchanged — that one is not a guess. xterm.js drops the extended record whenever the underline style is `NONE` and both getters fall back to reporting the foreground, so `ulColor === fg && ulMode === fgMode` is the shape of "no record", and `underline_color: None` is already the vocabulary's spelling for an underline that follows the text color.

crates/tui-test/assets/xtermjs/shim.js
aymanbagabas: Added in 0fc1fbb, as conformance cases rather than xterm.js-only tests so they hold every backend to the same contract:

- **`conformance_resetting_the_underline_color_is_not_white`** covers the `SGR 59` case from your comment below: set red, reset, then set a genuine white, and assert all three read back distinctly. Verified it fails on xterm.js with the old read-time heuristic restored and passes on alacritty, so it is a real guard rather than a test written around the current behaviour.

- **`conformance_default_and_styled_cells_do_not_blur_together`** covers the `isAttributeDefault()` fast path — the other special case in `pack`, where one call stands in for nineteen getters. It asserts a default cell before and after a heavily styled one reports nothing, and that the styled cell keeps fg, bg, underline, bold, italic, inverse, and strike. That shortcut had only a comment claiming it was "verified to produce byte-identical output" with nothing checking it.

Both pass on alacritty, ghostty, rio, and xtermjs.

The dynamic-color special case has `conformance_one_dynamic_color_sequence_sets_successive_colors` and `conformance_each_dynamic_color_query_is_answered_separately` from your OSC 10-12 comment, so the shim's four notable divergences are now all covered.

The remaining shim special cases are the ones that supply something the bundle omits rather than correct something it gets wrong — window title, cursor visibility, cursor shape — and those are already exercised by the existing conformance cases for each.

crates/tui-test/src/terminal/xtermjs.rs
aymanbagabas: Both true, and the second one was the worse half — `process` and `resize` discarded the `Result` entirely, so a throw meant the grid silently stopped tracking the bytes and every later read answered from a stale grid as if it were fact.

Fixed in 0fc1fbb. Every call now goes through one helper that records the exception instead of substituting the call's own default:

```rust
fn invoke<R, F>(&self, method: &str, body: F) -> Option<R> { ... }
```

The message is claimed with `ctx.catch()` and carries the JS stack. rquickjs reports a throw as a bare `Error::Exception` and parks the value on the context, so it has to be taken there or the next call overwrites it.

Surfacing it needed a route out, since `Emulator` is infallible by design and the reader thread has nobody to return an error to. So:

- `Emulator::fault() -> Option<String>` defaults to `None`. Backends that parse in-process cannot fail this way; only one driving a separate engine can.
- `XtermJsEmu` keeps the first fault. Once the engine has thrown, what follows is a consequence rather than a new fact, and the earliest message names the cause.
- `Engine::with_session` checks it **before** running the operation and returns `TuiTestError::internal`. Every session operation routes through there, so one check covers all of them. Before rather than after on purpose: once the grid has stopped tracking, a wrong answer is worse than a failure.

`a_thrown_exception_is_reported_rather_than_swallowed` covers it — replaces `feed` with a thrower the way a bundle upgrade moving a private API would, then asserts the message reaches the caller and that a later failure does not overwrite it.

Worth noting this is not hypothetical: the shim reads several things xterm.js does not expose publicly (`coreService.isCursorHidden`, `decPrivateModes.cursorStyle`, `_inputHandler._curAttrData`), so a release moving one is exactly how this fires. Previously that would have shown up as an inexplicably empty screen.

crates/shell-use/assets/xterm/README.md
aymanbagabas: Reopening this one — I've reverted the de-vendoring in 8ff376c, so the bundles are committed again. Flagging rather than quietly changing it, since it goes against what you asked for above.

The reason is the one cost I did not weigh properly when I made the change: `build.rs` put npm and a network on the path of anyone building from a clone. Not just publishing or CI — a plain `cargo build --features xtermjs`, and by extension the CLI and both bindings, which enable the feature by default. That is a heavier requirement than it looked, and it is paid by every contributor on every clean checkout rather than once at release time.

What that bought was keeping ~264 KB out of git. What it cost was a build script, an explicit `include` list in `Cargo.toml` to put the fetched files back into the published crate, `--allow-dirty` on both publish steps, and a network dependency for contributors. Removing all four is the simplification.

Your underlying concern still holds though, and I think it is better served without the fetch:

- **`vendored.yml` is back.** It re-fetches both packages at the pinned versions whenever these files change and fails if what is committed is not byte for byte what npm publishes. That is what makes a vendored blob checkable rather than trusted — it closes the loop a checked-in hash cannot.
- **The release check you asked for stays**, so a release cannot ship on a stale pin.
- **The helper script stays and got better**: `.github/scripts/vendor-xtermjs.sh --latest` bumps `pinned.json` to the newest releases and re-fetches in one step, so updating is one command rather than the block of `npm pack`/`tar` that used to be in the README.

So the delta versus what you asked for is only where the bytes live. Everything about drift — stale pins, edited blobs, provenance — is enforced by CI either way.

Happy to go back to fetching if you would still rather not have them in the tree; the implementation is in 0fc1fbb and reverting the revert is easy. I would want to move the feature off default-on for the CLI and bindings first, though, so the network requirement does not land on people who never asked for the xterm.js backend.

.github/workflows/xtermjs-update.yml
aymanbagabas: One correction to the above: I said I had dropped `.github/workflows/vendored.yml`. That was tied to de-vendoring, which I have since reverted in 8ff376c — the bundles are committed again, so that workflow is back and meaningful again. Details in the vendoring thread.

Everything else in this comment still holds: the weekly-PR workflow is gone, and the release `verify` job fails unless both pinned versions are the latest published ones.

crates/shell-use/assets/xterm/README.md
aymanbagabas: Correction to my last reply: `vendored.yml` is gone again (5be93d7). Re-vendoring is a manual step, so nothing in CI reaches for the tarballs — `.github/scripts/vendor-xtermjs.sh` is the only thing that rewrites those files.

What remains, and the distinction that matters:

- **`.github/scripts/vendor-xtermjs.sh --latest`** — manual, bumps `pinned.json` to the newest releases and re-fetches in one command.
- **The release `verify` check** — compares version strings via `npm view` and never touches the bundles, so a release cannot ship on a stale pin. That is the step you asked for in the other thread; it is unaffected.

So provenance is established when a human vendors, and recorded in `pinned.json` and the commit, rather than re-asserted by CI on every push.

.github/workflows/xtermjs-update.yml
aymanbagabas: Further correction on `vendored.yml`: it is removed again in 5be93d7, and this time for good — re-vendoring is manual, so nothing in CI should be re-fetching the tarballs.

The step you asked for here is unaffected. It compares version strings with `npm view` and never touches the bundles, so it still fails a release whose pins are behind the latest published versions. I have made that explicit in a comment above it so the two do not get conflated again.

crates/tui-test/assets/xtermjs/README.md
aymanbagabas: Small correction to the numbers above: it is 114 lines down to 74, not 118 to 81 — I quoted the counts before the last couple of edits landed. The substance is unchanged.

.github/workflows/xtermjs-update.yml
aymanbagabas: Reopening — I've gone back to the weekly updater in 2efdd91 and removed the release-time check. Flagging rather than quietly changing it, since it is the opposite of what you asked for here.

The reason is where the release check lands when it fires. It gates the release `verify` job, so an upstream publish — someone else's release, on their schedule — turns an otherwise-ready tag into a failed release. The bundle being one version behind is not a reason a release is unsafe, and coupling those two means the fix has to happen under time pressure at the worst moment.

The weekly PR moves that to a moment where it costs nothing, and it carries something the version check could not:

- It runs the **xterm.js conformance suite against the new bundles** before opening the PR. Nothing else does — CI builds with default features, so those cases never run on a runner. The shim reads several private xterm.js fields, so a release moving one is the failure mode that matters, and a version-string comparison cannot see it.
- The PR body flags the two things conformance does not cover: the emoji width table, and each package's separate `LICENSE`.

On the noise: it only opens a PR when there is actually a newer release, and it decides that from the vendored files rather than `pinned.json` alone, so a package republished at the same version is caught too. `@xterm/headless` has published roughly a couple of times a year, so in practice this is a PR every few months, not weekly noise. If it does turn out noisy, the cron is one line to slow down.

It is also much smaller than the version it replaced — it just calls `.github/scripts/vendor-xtermjs.sh --latest`, which is the same command a human runs, rather than reimplementing the `npm pack` / `tar` steps inline.

Manual bumping still works exactly as you described, and is what the README documents first.

Happy to go back to your version if you would rather have the release gate — or to keep both, if the objection was the weekly PR rather than the release coupling. The version-check step is in 5be93d7 if we want it back.

aymanbagabas: Follow-up: the release job does gate on the bundles again in bcb3c650, just on a different property than you proposed — provenance rather than freshness.

It re-fetches the versions `pinned.json` names and fails if the committed bundles are not byte for byte what npm publishes:

```
Both bundles match @xterm/headless@6.0.0 and @xterm/addon-unicode11@0.9.0.
```

The distinction that makes this safe to block a release on: it can only fail on something under our control — an edited blob, or a pin bumped without re-vendoring — so an upstream publish cannot turn a ready tag into a failed release. Freshness is the weekly job's business, and that opens a pull request rather than blocking anything.

Exercised both failure modes against the real packages rather than assuming:

```
edited blob        → xterm-headless.js is not @xterm/headless@6.0.0.
pin bumped, stale  → xterm-headless.js is not @xterm/headless@5.5.0.
                     Re-run .github/scripts/vendor-xtermjs.sh to restore them.
```

So the split is: the weekly PR notices a new release and proves the shim still works against it, and the release job proves the bytes are what they claim to be.

aymanbagabas: Scratch that — I've reverted it in 50b5d7e. It was fluff.

It re-asserted at release time something already settled when the files were written, for a mismatch that cannot reach the artifact: `pinned.json` is metadata, and the committed bytes are what compile in. A pin disagreeing with them is a wrong note, not a wrong build.

Neither failure mode I claimed it caught holds up:

- A hand-edited bundle is a diff against a minified file sitting in code review.
- A pin bumped without re-vendoring is what the script and the weekly job exist to prevent, and the weekly job overwrites both on its next run anyway.

Against that it put a network dependency and a Node toolchain on the release path, and could block a ready tag on a metadata inconsistency — which is the same objection that removed the freshness gate, only rarer.

Sorry for the churn in this thread. Where it has landed:

- **Weekly** — `.github/workflows/xtermjs-update.yml` opens a pull request when either package has a newer release, and runs the xterm.js conformance suite against the new bundles first, since nothing else does.
- **Manual** — `.github/scripts/vendor-xtermjs.sh --latest`, which is the same command the workflow runs.
- **Release** — nothing xterm.js-specific.

Still leaving this unresolved: the weekly PR is what you argued against, so whether it stays is your call.

crates/shell-use/assets/xterm/README.md
aymanbagabas: Resolving this to keep the review tidy — not to close the question. Where it landed, for the record:

**The bundles stay committed.** `crates/tui-test/assets/xtermjs/{xterm-headless.js,addon-unicode11.js}`, 264 KB, byte-for-byte what npm publishes at the versions `pinned.json` names.

The deciding factor was that fetching at build time put npm and a network on the path of anyone building from a clone — not just publishing or CI, but a plain `cargo build`, since the CLI and both bindings enable the feature by default. That is paid by every contributor on every clean checkout, to keep 264 KB out of git.

Provenance is reproducible rather than trusted: `.github/scripts/vendor-xtermjs.sh` regenerates both files from npm, and re-running it on the current pins is a no-op, which is the property that makes the committed bytes checkable by anyone at any time.

The de-vendored implementation is in 0fc1fbb if you want it back — it worked, `cargo publish --dry-run` and an npm-free build both verified. If we do revisit it, I would move the feature off default-on for the CLI and bindings first, so the network requirement does not land on people who never asked for the xterm.js backend.

Happy to reopen if you would rather go the other way.

.github/workflows/xtermjs-update.yml
aymanbagabas: Resolving this to keep the review tidy — not to close the question. Where it landed:

**Weekly pull request, no release gate.** `.github/workflows/xtermjs-update.yml` runs `.github/scripts/vendor-xtermjs.sh --latest` on a cron, and opens a PR only when either package actually has a newer release. Nothing xterm.js-specific runs at release time.

Two things pushed it this way. A release gate fires on someone else's schedule — an upstream publish turns an otherwise-ready tag into a failed release, and a bundle being one version behind is not a reason a release is unsafe. And a version-string comparison cannot see the failure that actually matters: the shim reads several private xterm.js fields, so a release moving one is the real risk, and the weekly job runs the conformance suite against the new bundles before opening the PR. Nothing else does, since CI builds with default features.

On the noise you were worried about: it opens a PR only on a real new release, and decides that from the vendored bytes rather than `pinned.json` alone, so a package republished at the same version is caught too. `@xterm/headless` publishes roughly a couple of times a year. If it does turn out noisy, the cron is one line.

Manual bumping still works exactly as you described, with the same command the workflow runs, and it is what the README documents first.

I also tried and then reverted a release-time provenance check (bcb3c650, reverted in 50b5d7e) — it re-asserted something already settled when the files were written, for a mismatch that could not reach the artifact. Sorry for the churn here.

The version-check step is in 5be93d7 if you want it back, and I am happy to reopen.

cpendery (approved):

PR #84, opened 2026-08-04
fix: unify colour resolution behind a terminal profile

aymanbagabas:
First of a three PR stack: **#84 → #85 → #86**.

A screenshot and a color assertion disagreed about what color a cell was, because each resolved palette indices through its own hardcoded table. `render/svg.rs` had a private `Theme` where red was `#e88388`; `assert/color.rs` had an `ANSI16` table where red was `#800000`. Nothing kept them in sync and nothing noticed they had drifted.

### The bug

Same program, same red text, on `main`:

```console
$ shell-use submit 'printf "\033[31mERROR\033[0m: red text\n"'
$ shell-use screenshot --out shot.svg     # paints ERROR #e88388

$ shell-use expect text ERROR --fg '#800000'   # a color not in the image
$ echo $?
0                                              # ...passes

$ shell-use expect text ERROR --fg '#e88388'   # the color you can see
$ echo $?
1                                              # ...fails
```

Asserting the color the image actually paints **failed**, and asserting a color that appears nowhere **passed**. Either the picture was lying or the assertion was, and a test suite could not tell you which.

| before — image paints `#e88388` | after — image paints `#800000` |
| --- | --- |
| before | after |

Both tables are deleted. One `Colors` type resolves every index once, so the renderer and the assertion cannot disagree by construction. `--fg '#800000'` now passes *and* matches the picture.

### Profiles

The palette is no longer a private constant, so a session can be given one. `--config` picks the file and `--profile` the entry; discovery cascades `./shell-use.toml` then `~/.shell-use/shell-use.toml`.

```toml
[profiles.solarized]
scrollback = 10000

[profiles.solarized.colors]
background = "#002b36"
foreground = "#839496"
cursor     = "#d33682"
red        = "#dc322f"
green      = "#859900"
blue       = "#268bd2"
```

```console
$ shell-use run --config shell-use.toml --profile solarized -- bash
```

solarized

A profile is exactly 19 values: the 16 ANSI colors plus foreground, background, and cursor. It is read only for the life of the session, so a program cannot rewrite the palette a test was pinned against. Anything outside 0-15 comes from a static xterm table, and `scrollback` defaults to 10000.

### Notes for review

- `--fg <index>` compares the cell's palette *index*, so it was never affected by this bug and still is not. Only `#rrggbb` resolves through the palette.
- Snapshots record the slot (`"fg": 1`), not RGB, so existing baselines survive a palette change. There are tests pinning that.
- Defaults are the VGA palette, foreground `#c0c0c0` on background `#000000`, which is what the assertion side already used, so no assertion changes meaning.


aymanbagabas:
@cpendery could you take a look? This is the base of a three PR stack:

- **#84** (this one) — one palette instead of two that disagreed, plus configurable profiles
- #85 — answer OSC color queries, and support setting/resetting colors at runtime
- #86 — draw the cursor in screenshots

Read them in that order. GitHub could not request you as a reviewer directly (I do not have push access here), hence the mention.

aymanbagabas:

aymanbagabas:
Reopened as #94, now the bottom of a real stack: #94 → #95 → #96.

PR #85, opened 2026-08-04
feat: answer OSC colour queries and track dynamic colours

aymanbagabas:
Second of a three PR stack: #84 → **#85** → #86. Stacked on #84; a cross-repo PR cannot be based on a fork branch, so GitHub shows #84's commit here too. Review only `feat(terminal): answer OSC color queries` onward.

A terminal is supposed to answer when a program asks what colors it is using. Ours never did, so every program that asked had to time out and guess.

### Before: the query goes unanswered

A program asking for the background (`OSC 11`) and cursor (`OSC 12`) color:

| before — no reply, twice | after — answered |
| --- | --- |
| before | after |

That silence is not harmless. Programs use `OSC 11` to decide whether they are on a light or dark background, so a timeout means an editor or a diff tool picks a theme by guessing, and an agent driving the terminal sees whichever colors the guess produced rather than the ones actually configured. It also costs a real second of wall clock per query while the program waits.

### After: setting works too

`OSC 10`, `OSC 11`, `OSC 12` set the foreground, background, and cursor; `OSC 4` sets a palette entry. `OSC 110`, `OSC 111`, `OSC 112`, and `OSC 104` reset them.

```console
$ printf '\033]11;#1c2833\a'      # background
$ printf '\033]10;#eaeaea\a'      # foreground
$ printf '\033]4;1;#ff5f87\a'     # palette entry 1, red

$ printf '\033]111\a\033]110\a\033]104;1\a'   # put them all back
```

| default | after the three sets | after the three resets |
| --- | --- | --- |
| default | set | reset |

Background `#000000` → `#1c2833` → `#000000`, red `#800000` → `#ff5f87` → `#800000`. The reset returns to the profile from #84, not to a second hardcoded default.

### How it resolves

```
emulator.color(slot) = what a program set with OSC   (runtime, clearable)
                    ?? the session profile           (from #84, read only)
                    ?? the static xterm table        (indices 16-255)
```

A reset clears the runtime value only. It can never clear a configured one, so `OSC 104` from a stray program cannot wipe the palette a test was pinned against.

### Notes for review

- The reply is built from alacritty's `Event::ColorRequest`, whose formatter already captured the query's prefix and terminator, so a `BEL`-terminated query gets a `BEL`-terminated answer and an `ST`-terminated one gets `ST`. There is no second parser over the PTY stream, and `Term::colors()` is readable directly once `advance()` returns.
- Replies are queued into the same pending buffer as `PtyWrite`, so ordering with other terminal output is unchanged.
- `ColorSlot` is an enum (`Indexed(u8) | Foreground | Background | Cursor`). Alacritty's internal 256/257/258 numbering stays inside `alacritty.rs` and does not leak into the profile type.
- 22 tests, including conformance cases that run against every backend, so a future backend inherits them.


aymanbagabas:
@cpendery second of the stack, on top of #84 — please review that one first. Only `feat(terminal): answer OSC color queries` onward is new here; the base commit shows up because a cross-repo PR cannot be based on a fork branch.

aymanbagabas:

aymanbagabas:
Reopened as #95, based on #94 so the diff shows only this layer.

PR #91, opened 2026-08-05
feat(terminal): track, draw, and assert on the window title

aymanbagabas:
A program announces what it is doing by setting the window title with `OSC 0` or `OSC 2`, and shell-use threw that away. An agent driving a terminal could see what a program *printed* but not what it *called itself* — which is exactly where long-running tools report progress, and how a shell says which command is running.

The emulator now tracks the title, and it is reachable everywhere the other terminal properties already are.

### Before / after

Same session, same command, screenshotted by the same `screenshot --out`:

| before | after |
| --- | --- |
| before | after |

The title is drawn centred in the title bar, clamped so it can never run under the window controls, and truncated with an ellipsis when it does not fit. No title leaves the bar exactly as it was, so every screenshot taken without one is byte-identical to before.

### Surfaces

```console
$ shell-use get title
{ "value": "vim: notes.md" }

$ shell-use expect title "vim"                  # exit 0
$ shell-use expect title "emacs"                # exit 1
$ shell-use wait title "build complete" --timeout 60000
$ shell-use expect title '^\w+: \S+\.md$' --regex
$ shell-use state | grep title
title: vim: notes.md
```

`state.title`, `get title`, `expect title`, and `wait title`, plus `get_title` / `wait_title` / `expect_title` in the Python and JS bindings. All four take `--regex`, `--not`, and `--timeout` where the sibling text commands do.

### Widths are counted in columns, not characters

A terminal lays out by column. A CJK glyph is one `char` but two columns; a combining mark is one `char` but none. Counting characters drew the snapshot frame four columns out of true, and in the centred title bar it sized the title at half its real width so it spilled over the window controls at both ends.

cjk

```
╭─ 编辑：第三章笔记 ──────────────────╮      36 columns
│ok                                │      36 columns
╰──────────────────────────────────╯      36 columns
```

One shared `truncate_to_columns` in `terminal/cell.rs` serves both. This is the only reason `unicode-width` is now named in `Cargo.toml`: `alacritty_terminal` already pulls the identical 0.2.2, so declaring it added **one line to `Cargo.lock` and zero new packages**. Neither `alacritty_terminal` nor `vte` re-exports it, and hand-rolling East-Asian-width tables would be both reinvention and wrong on emoji and combining marks — the tests cover `🚀 build` and `e\u{301}clair` for exactly that reason.

### An empty title means reset

Programs clear the title by sending an empty one, so `OSC 2;` reports as `None` rather than as a title that happens to be blank. Keeping the two apart would only hand every caller the same special case to write.

### Snapshots record the title only when asked

`expect snapshot --include-title` puts it in the frame's top border; the default leaves it out.

This is deliberate and worth a look during review. A stock Ubuntu `.bashrc` sets `PS1='\[\e]0;\u@\h: \w\a\]…'`, so a normal shell writes its own title on every prompt. Recorded by default, that is:

```
╭─ ayman@Mac-91: /tmp/ttl ─────────╮
```

which pins the baseline to one username, hostname, and absolute path, and changes on `cd` while the screen stays identical. Off by default, a snapshot with no title is byte-identical to one taken before this PR, so every stored baseline keeps passing.

### Free

The title stack (`CSI 22 t` push, `CSI 23 t` pop) works, because alacritty implements a pop as setting the title it popped and that routes through the same event. There is a test for it rather than any code.

### Notes for review

- The title is tracked from `Event::Title` / `Event::ResetTitle` in the existing `CaptureProxy` rather than read back off the terminal, because alacritty exposes no way to read it: `Term.title` and `Term.colors` are both private fields, but only `colors` has a public getter (`pub fn colors(&self) -> &Colors`). There is no `Term::title()`, and `set_title` is a `Handler` setter. Alacritty's own frontend consumes the same events for the same reason.
- `OSC 1` (icon name) is deliberately not supported: vte never dispatches it, and with a single drawn title there is nothing for a separate icon name to do.
- The title is markup-escaped in the SVG like any other text; it is whatever the program chose to send, and there is a test that a title of `</text><script>` cannot inject an element.
- Tests: 5 conformance cases (which run against every backend, so a future backend inherits them), 5 renderer tests, 4 snapshot tests, and 2 end-to-end tests through a real PTY.


aymanbagabas:
@cpendery ready for review when you have a moment. Branches off latest `main`, standalone (not part of the #84 → #85 → #86 stack).

GitHub will not let me request you as a reviewer directly (no push access here), hence the mention.

crates/shell-use/src/assert/snapshot.rs
```diff
@@ -101,9 +106,30 @@ pub fn serialize(rows: &[Vec<EmuCell>], cols: u16, include_colors: bool) -> Stri
...
+/// long for the border is truncated so the frame stays rectangular.
+fn box_view(view: &str, width: u16, title: Option<&str>) -> String {
+    let width = width as usize;
+    let bar = "─".repeat(width);
+    // `╭─ title ───╮`: one leading dash, the spaced title, then at least one
+    // trailing dash. A title with no room for even one character is dropped
+    // rather than allowed to push the corner out of line.
+    let label = title.and_then(|title| {
+        let room = width.checked_sub(4).filter(|room| *room > 0)?;
+        Some(format!(" {} ", truncate_to_columns(title, room)))
```
cpendery, line 140: are we sure we want to truncate the title for the snapshot boxes? it might make more sense to expand the `{view}\n{attributes}` such that the attributes could be `{title: "title", colors: {}, }` and so on.

maybe we only drop into the attributes it the title would be truncated

crates/shell-use/src/terminal/cell.rs
```diff
@@ -7,6 +7,7 @@
 
 use bitflags::bitflags;
 use compact_str::CompactString;
+use unicode_width::UnicodeWidthChar;
```
cpendery: I think you'll run into some weird width calculation issues using this, maybe UnicodeWidthStr is a better option

crates/shell-use/src/engine.rs
```diff
@@ -1139,10 +1279,15 @@ fn do_snapshot(
...
     include_colors: bool,
+    include_title: bool,
     cwd: Option<String>,
 ) -> Result<SnapshotResult, ShellUseError> {
     let rows = viewable(session);
-    let content = snapshot::serialize(&rows, session.cols, include_colors);
+    // Off by default: a shell prompt routinely sets the title to a username,
+    // hostname, and absolute path, which would pin every baseline to one
+    // machine and make it change on `cd` while the screen stayed the same.
+    let title = include_title.then(|| title_of(session)).flatten();
```
cpendery: nit: might have a weird race case here where we release the title lock on the session, the pty continues, and then we snapshot the grid. should flag & address in a follow up pr

cpendery (commented):
nit: can we also update the bindings readmes with the new methods?

crates/shell-use/src/terminal/cell.rs
aymanbagabas: Good catch, and it is worse than a nit. Measured on 0.2.2:

| sequence | per character | `UnicodeWidthStr` |
| --- | --- | --- |
| `👨‍👩‍👧‍👦` | 8 | 2 |
| `👍🏽` | 4 | 2 |
| `1️⃣` | 1 | 2 |
| `❤️` | 1 | 2 |

Switched to `UnicodeWidthStr`. Truncation now finds its cut by measuring real prefixes rather than accumulating per-character widths, so the result is exactly as wide as it measured even when the cut lands inside a sequence. Tests cover all four cases above plus every budget from 0 to 11.

One thing this does not fix: the grid itself measures per character, because `alacritty_terminal` uses `UnicodeWidthChar`. So a *row* holding one of these already breaks the box today, with no title involved — filed as #92, since fixing it is a decision about the emulator seam rather than the serializer.

crates/shell-use/src/assert/snapshot.rs
aymanbagabas: You are right that truncating alone is wrong here. A snapshot is an assertion, not a picture: `building module A, step 3` and `building module B, step 7` both shorten to `building module A, …` in a narrow frame, so they would record identically and pass for each other.

Taken your suggestion, with the border kept as well. The border still shows a shortened title because that is the readable part, and the exact one is recorded in the attributes:

```
╭─ building module A, … ─╮
│ok                      │
│                        │
╰────────────────────────╯
{
  "title": "building module A, step 3"
}
```

Recorded unconditionally rather than only when truncated, so the shape does not depend on terminal width. Attributes are now the object you suggested, with the shifts under `colors`, which does mean a stored snapshot taken with `--include-colors` needs re-recording once. No baselines in the repo, and `--include-title` is still off by default, so a plain snapshot is unchanged.

crates/shell-use/src/engine.rs
aymanbagabas: Real race, so I fixed it here rather than deferring — it is the exact flakiness `--include-title` being opt-in was meant to avoid, and a shell writes its prompt and *then* sets its title, so the window is not theoretical.

There were three sites, not one: `do_snapshot`, `assertion_message`, and `screenshot` all took the grid under one lock and the title under another. They now share a `grid_with_title` helper that reads both under a single guard.

Happy to split it out if you would rather review it separately.

aymanbagabas:
Thanks @cpendery — all four addressed in 996c5f9, replies inline on each thread.

- **Bindings READMEs** — `get_title`/`getTitle`, `wait_title`/`waitTitle`, `expect_title`/`expectTitle` added to both API lists.
- **`UnicodeWidthStr`** — adopted; it differed on 4 of 10 cases I measured. Also surfaced a pre-existing bug where rows holding an emoji sequence already break the box, filed as #92.
- **Snapshot title** — border keeps a shortened title, attributes now record the exact one so two long titles cannot collapse into one baseline. Note this nests colour shifts under `colors`, so a `--include-colors` baseline needs re-recording once.
- **Lock race** — fixed here rather than deferred; there were three sites, now sharing one helper.

cpendery (approved):

PR #93, opened 2026-08-05
feat: track and assert on bell

cpendery:


bindings/js/native/index.d.ts
```diff
@@ -198,6 +201,7 @@ export interface State {
   last_exit: number | null
   exited: number | null
   ready: boolean
+  bell_count: number
```
aymanbagabas, line 236: Do we want to timestamp them? I think we can do much more if we have timestamps

bindings/js/native/index.d.ts
```diff
@@ -16,6 +16,7 @@ export declare class NativeSession {
   getCwd(): Promise<string | null>
   getCursor(): Promise<Cursor>
   getSize(): Promise<Size>
+  getBellCount(): Promise<number>
```
aymanbagabas, line 19: Shouldn't we change the bell API to return the recorded bell events instead of just a number? `getBellEvents(): Promise<BellEvent[]>`

PR #94, opened 2026-08-05
fix: unify colour resolution behind a terminal profile

aymanbagabas:
First of a three PR stack: **#94** → #95 → #96.

A screenshot and a color assertion disagreed about what color a cell was, because each resolved palette indices through its own hardcoded table. `render/svg.rs` had a private `Theme` where red was `#e88388`; `assert/color.rs` had an `ANSI16` table where red was `#800000`. Nothing kept them in sync and nothing noticed they had drifted.

### The bug

Same program, same red text, on `main`:

```console
$ shell-use submit 'printf "\033[31mERROR\033[0m: red text\n"'
$ shell-use screenshot --out shot.svg     # paints ERROR #e88388

$ shell-use expect text ERROR --fg '#800000'   # a color not in the image
$ echo $?
0                                              # ...passes

$ shell-use expect text ERROR --fg '#e88388'   # the color you can see
$ echo $?
1                                              # ...fails
```

Asserting the color the image actually paints **failed**, and asserting a color that appears nowhere **passed**. Either the picture was lying or the assertion was, and a test suite could not tell you which.

| before — image paints `#e88388` | after — image paints `#800000` |
| --- | --- |
| before | after |

Both tables are deleted. One `Colors` type resolves every index once, so the renderer and the assertion cannot disagree by construction. `--fg '#800000'` now passes *and* matches the picture.

### Profiles

The palette is no longer a private constant, so a session can be given one. `--config` picks the file and `--profile` the entry; discovery cascades `./shell-use.toml` then `~/.shell-use/shell-use.toml`.

```toml
[profiles.solarized]
scrollback = 10000

[profiles.solarized.colors]
background = "#002b36"
foreground = "#839496"
cursor     = "#d33682"
red        = "#dc322f"
green      = "#959900"
blue       = "#268bd2"
```

```console
$ shell-use run --config shell-use.toml --profile solarized -- bash
```

solarized

A profile is exactly 19 values: the 16 ANSI colors plus foreground, background, and cursor. It is read only for the life of the session, so a program cannot rewrite the palette a test was pinned against. Anything outside 0-15 comes from a static xterm table, and `scrollback` defaults to 10000.

### Notes for review

- `--fg <index>` compares the cell's palette *index*, so it was never affected by this bug and still is not. Only `#rrggbb` resolves through the palette.
- Snapshots record the slot (`"fg": 1`), not RGB, so existing baselines survive a palette change. There are tests pinning that.
- Defaults are the VGA palette, foreground `#c0c0c0` on background `#000000`, which is what the assertion side already used, so no assertion changes meaning.


SKILL.md
```diff
@@ -60,7 +60,7 @@ without parsing text:
 
 | Command                                                                  | Description                                                            |
 | ------------------------------------------------------------------------ | ---------------------------------------------------------------------- |
-| `open [--shell S] [--cols N] [--rows N] [--cwd D] [--env K=V]...`        | Spawn a shell session (auto-starts the daemon). `--env` is repeatable. |
+| `open [--shell S] [--cols N] [--rows N] [--cwd D] [--env K=V] [--config F] [--profile P]...` | Spawn a shell session (auto-starts the daemon). `--env` is repeatable. |
```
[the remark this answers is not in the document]
aymanbagabas: Right, and it is my slip: the ellipsis belonged to `--env`, and appending `--config` and `--profile` after it moved the mark onto a flag that takes a single name. Reordered so it trails `--env` again, matching the `run` line just below.

The README line has no ellipsis, so it was unaffected.

crates/tui-test/src/profile.rs
```diff
@@ -0,0 +1,522 @@
...
+                    }
+                };
+                Rgb::new(level((i / 36) % 6), level((i / 6) % 6), level(i % 6))
+            }
+            232..=255 => {
+                let v = ((index as u16 - 232) * 10 + 8) as u8;
+                Rgb::new(v, v, v)
+            }
+        }
+    }
```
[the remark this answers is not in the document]
aymanbagabas: Measured this rather than guessing, and I am leaving it as is — happy to change it if you disagree.

Benchmark of the two shapes, release, 4000 lookups (an 80x24 screenshot resolves roughly that many):

| | per lookup | per screenshot |
| --- | --- | --- |
| `ansi()[i]` | 5.9 ns | 23.5 µs |
| 16-arm match | 1.3 ns | 5.3 µs |

So the array is ~4.5x slower per lookup but the whole saving is **18 µs per screenshot**, against writing thousands of formatted SVG elements and a file. Adding `#[inline]` to `ansi()` does not close it (5.9 → 5.1 ns) because the index is runtime-variable, so the array still has to be materialised. The fix would be sixteen lines of match to buy that.

One correction to the premise, though: on the next PR in the stack (#95) `Colors::rgb` stops being the resolution path entirely — it is only reachable from tests there, and the live path becomes `AlacrittyEmu::color`, which reads `colors.ansi()` directly. So if this is worth doing it is worth doing there, not here.

aymanbagabas:
Review pass on the stack — validated each comment before touching anything, and two of the six turned out not to need a change.

**Fixed here:** the `--env` ellipsis, which my own edit had shifted onto `--profile`.

**Not changed, thread left open for you:** the `ansi()` array in `Colors::rgb`. Benchmarked at 5.9 ns vs 1.3 ns per lookup, which is **18 µs per screenshot** — sixteen lines of match to buy that. Details on the thread, and note `rgb()` stops being the live path in #95 anyway.

The other four are on #95 and #96, two of which were real bugs:
- **colour replies were reordered** — a query answered *after* a device attributes reply reads as unanswered and then leaks into input (#95)
- **cursor row truncated at 65535** — `scrollback = 70000` drew the cursor on row 2467 instead of 68002 (#96)

README.md
```diff
@@ -299,6 +299,54 @@ Every command returns a stable exit code so an agent can branch on the failure c
...
+scrollback = 500              # inherits the default palette
+```
+
+```bash
+shell-use open                         # profile "default"
+shell-use open --profile ci
+shell-use open --config ./other.toml --profile ci
+```
+
+Looked up nearest first: `./shell-use.toml`, then
```
cpendery: I think we should also honor XDG_CACHE_HOME if we are adding a configuration file

crates/tui-test-cli/src/cli.rs
```diff
@@ -34,6 +34,28 @@ impl From<ShellArg> for Shell {
     }
 }
 
+/// Which terminal profile a session runs with.
+#[derive(Args, Clone, Default)]
+pub struct ProfileArgs {
+    /// Config file to read (default: ./shell-use.toml, then
+    /// ~/.shell-use/shell-use.toml).
+    #[arg(long, value_name = "PATH")]
+    pub config: Option<std::path::PathBuf>,
```
cpendery, line 43: If we are adding a config, we should add some validation to it, ex: validating that we are getting valid colors for the profile so we don't have random crashes

cpendery (commented):
Overall looks good, one gap to call out is the handling of dim / inverse. The SVG renderer still handles these properly, but the assertions are just based on the raw SGR state. 

We should at least use the transformed color for the assertions. Additionally, adding `--dim` / `--inverse` flags would allow for easier assertions for the end user on those transformed colors

crates/tui-test-cli/src/cli.rs
cpendery: adding as a follow up issue

cpendery (approved):

PR #100, opened 2026-08-06
ci: add cargo publishing

cpendery:


.github/scripts/release/publish-crate.mjs
aymanbagabas, line 1: Why do we need this? Can't we just `cargo publish ...` things with the right token permissions?

.github/workflows/release.yml
```diff
@@ -29,13 +29,20 @@ jobs:
...
+
       - uses: actions/setup-node@v4
         with:
           node-version: "24"
 
       - name: Verify release versions
         run: node .github/scripts/release/verify-versions.mjs
 
+      - name: Verify shell-use crate package
+        run: node .github/scripts/release/publish-crate.mjs --dry-run
```
aymanbagabas: Ditto

.github/workflows/release.yml
```diff
@@ -358,6 +365,41 @@ jobs:
...
+        with:
+          ref: ${{ env.RELEASE_TAG }}
+
+      - run: |
+          rustup toolchain install stable --profile default
+          rustup default stable
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "24"
```
aymanbagabas: ditto

cpendery: the auth stage is required because I setup trusted publishing for the pipeline, https://crates.io/docs/trusted-publishing

.github/workflows/release.yml
cpendery: I think it's useful to do dry runs for all the packaging before we start publishing

.github/scripts/release/publish-crate.mjs
cpendery: removed a lot of the script to just include publishing in the pipeline

aymanbagabas (approved):

PR #132, opened 2026-08-11
feat: add kitty keyboard support

cpendery:


crates/tui-test/src/input/keys.rs
```diff
@@ -1,308 +1,624 @@
...
+    let mut mods = Mods::default();
+    let mut event = None;
+    for prefix in prefixes {
+        match prefix.to_ascii_lowercase().as_str() {
+            "ctrl" | "control" => mods.ctrl = true,
+            "alt" | "option" | "meta" => mods.alt = true,
+            "shift" => mods.shift = true,
+            "super" | "command" | "cmd" | "win" | "windows" => mods.super_key = true,
+            "hyper" => mods.hyper = true,
+            "kittymeta" | "kitty_meta" => mods.meta = true,
```
aymanbagabas: What is `kittymeta`? We should really have `mods.alt` and `mods.meta`. Some environments like X11 and macOS differentiate them. combining all of `"alt" | "option" | "meta"` doesn't seem right.

README.md
```diff
@@ -221,11 +221,13 @@ print the screen bare.
 | ------------------------------------------------------------- | ------------------------------------------------------------ |
 | `type "text"`                                                 | Type literal text.                                           |
 | `submit ["text"]`                                             | Type then press the shell return key.                        |
-| `press <Key...>`                                              | Named keys, e.g. `press Escape : w q Enter`, `press Ctrl+C`. |
-| `keys "Control+a"`                                            | A single key combo.                                          |
+| `press <Key...>`                                              | Named keys and events, e.g. `press Ctrl+C`, `press Repeat+Up`, `press Release+a`. |
+| `keys "Ctrl+a"`                                               | A single key combo or event.                                 |
```
aymanbagabas: Why don't we have `press`, `release`, and `repeat` as commands instead? We can fold `keys` into these commands. There's also `mouse release` that is kind of related.

crates/tui-test/src/input/keys.rs
cpendery: I changed the options to the mapping below. I had originally used Meta as an alias for alt, so I was using kittymeta to map to the actual mod.meta
```rs
"alt" | "option" => mods.alt = true,
"meta" => mods.meta = true,
"super" | "cmd" | "command" | "win" => mods.super_key = true,
```

README.md
cpendery: I updated the apis to use a similar api to playwright with `terminal.keyboard.press("")` or `termina.keyboard.down("")`

the cli now has `key press | key down | key up | key repeat`

bindings/js/native/index.d.ts
```diff
@@ -20,6 +20,9 @@ export declare class NativeSession {
   type(text: string): Promise<void>
   submit(data?: string | undefined | null): Promise<void>
   press(keys: Array<string>): Promise<void>
+  keyDown(keys: Array<string>): Promise<void>
+  repeat(keys: Array<string>): Promise<void>
+  keyUp(keys: Array<string>): Promise<void>
```
aymanbagabas, line 27: I assume `press` and `keyDown` are the same here, right? If we want to go DRY then I would choose one naming convention over the other and remove redundant ones. Personally, I would go with `key press`, `key release`, `key repeat`

cpendery: I updated it to be more clear that press handles sending both a key down & key up event when reporting is active, similar to how "click" vs. "mousedown" & "mousedown"

Issue #179, opened 2026-08-26
xtermjs: enable the Kitty keyboard protocol once the vendored bundle reaches 6.1.0


aymanbagabas:
## Context

`XtermJsEmu` is currently the only backend that reports no Kitty keyboard flags. It declares this in its conformance opt-in:

```rust
// crates/tui-test/src/terminal/xtermjs.rs
crate::emulator_conformance_tests!(
    |cols, rows, profile| { ... },
    crate::terminal::conformance::Divergences {
        underline_color_needs_a_style: true,
        no_kitty_keyboard: true,
    }
);
```

That divergence is a version limit, not a design one.

## What changed upstream

xterm.js added the protocol behind a new `vtExtensions` terminal option:

```ts
export interface IVtExtensions {
  /**
   * Whether the [kitty keyboard protocol][0] (`CSI =|?|>|< u`) is enabled.
   * ... The default is false.
   */
  kittyKeyboard?: boolean;
  kittySgrBoldFaintControl?: boolean;
  win32InputMode?: boolean;
  colorSchemeQuery?: boolean;
}
```

(<https://github.com/xtermjs/xterm.js/blob/d3e32b344dfe7dd6015cff6a9aeaaeaeccdc2789/typings/xterm-headless.d.ts#L358>)

We are pinned to `@xterm/headless` **6.0.0** (`crates/tui-test/assets/xtermjs/pinned.json`), which has no `vtExtensions` at all. Verified against both bundles:

| symbol | 6.0.0 (vendored) | 6.1.0-beta.302 |
| --- | --- | --- |
| `vtExtensions` | absent | present |
| `kittyKeyboard` | absent | present |
| `win32InputMode` | absent | present |
| `colorSchemeQuery` | absent | present |

`vtExtensions` first appears in the 6.1.0 beta line. `npm dist-tags` still lists `latest` as 6.0.0, and `.github/scripts/vendor-xtermjs.sh --latest` follows `latest`, so this lands on its own once 6.1.0 ships.

## What to do when the bundle reaches 6.1.0

1. Re-vendor: `./.github/scripts/vendor-xtermjs.sh --latest`.
2. In `crates/tui-test/assets/xtermjs/shim.js`, pass the option through in `__boot`, driven by the profile rather than hardcoded:
   ```js
   vtExtensions: { kittyKeyboard: <profile.kitty_keyboard> },
   ```
[...]

PR #180, opened 2026-08-26
feat(terminal): make the Kitty keyboard protocol configurable

aymanbagabas:
Adds a `kitty_keyboard` profile setting (default on) and fixes a gap where the ghostty backend silently reported no Kitty keyboard flags at all.

## Why

Two separate problems, one subject.

**There was no way to test the legacy path.** A TUI that negotiates the Kitty protocol still has to work on the many terminals that lack it, and that fallback was unreachable from tui-test: every backend that implemented the protocol always had it on.

**ghostty never reported any flags.** `Emulator::keyboard_mode` has a default returning `KeyboardMode::empty()`, and `GhosttyEmu` never overrode it. So every key reaching a ghostty session was legacy-encoded no matter what the child had negotiated: `key press` would send `\x1b[A` to a program that had asked for `CSI u` event reporting. libghostty-vt exposed the state as `Terminal::kitty_keyboard_flags()` the whole time; it was simply never called.

## What

`Profile::kitty_keyboard`, default `true`, plumbed through `tui-test.toml`, the JS binding (`kittyKeyboard`), and the Python binding (`kitty_keyboard`).

Gating is native wherever the backend offers a switch:

| backend | how it is gated | child's `CSI ? u` query |
| --- | --- | --- |
| alacritty | `Config::kitty_keyboard` | unanswered, as on a terminal without the protocol |
| rio | at the read; rio-vt takes no configuration | still answered |
| ghostty | at the read; `Terminal` takes no configuration | still answered |
| xtermjs | n/a, no implementation to gate | n/a |

alacritty's switch suppresses the push, pop, set and report paths, so a disabled session is indistinguishable from a terminal that never implemented the protocol. rio and ghostty have no equivalent, so they are gated where the mode is read and their query replies still go out. The comments at each site say exactly this rather than implying parity.

## xterm.js

Reports no flags because the vendored bundle contains no implementation. Verified, not inferred: `@xterm/headless` 6.0.0 has no `vtExtensions`, and no handler for `CSI > u`, `CSI = u`, or `CSI < u`. That is a limit of the emulator rather than of the mapping onto it, which is what `Divergences` is for, so it is declared there as `no_kitty_keyboard`.

Support arrives upstream in the 6.1.0 beta line via `vtExtensions.kittyKeyboard`. Tracked in #179 with the steps to take once the vendored bundle gets there; the divergence flag is the one line to delete.

## Tests

Two new conformance cases, so the contract is pinned for every backend rather than for the one that happened to be implemented first:

- `conformance_kitty_keyboard_modes_are_pushed_and_popped` re-checks push, stack, and pop, with each of the five flags mapping to its own bit. Skipped for backends declaring `no_kitty_keyboard`.
- `conformance_kitty_keyboard_can_be_turned_off` checks that a disabled profile reports no flags whatever the child pushes. Runs on **all** backends including xtermjs, since "off" is a claim every backend can honor.

A backend that under-reports now fails a named test instead of quietly falling back to legacy keys. This case is what would have caught the ghostty gap.

Plus `a_profile_can_turn_the_kitty_keyboard_protocol_off` for config parsing, and a default assertion in `an_empty_config_yields_the_defaults`.

`cargo test --workspace --features tui-test-rs/ghostty,tui-test-rs/rio,tui-test-rs/xtermjs` gives 423 passed, 0 failed.

## Not in this PR

Routing key *encoding* to a backend's own encoder, and the related discovery that `keys.rs` models no DECCKM at all. Separate PR.


cpendery (commented):
So I understand that we are no respecting the child requesting input to be sent via kitty keyboard; however, we are answering yes to child pty's with rio (& ghostty) that we support kitty keyboard. I think this could cause some weird failures in the future for apps that get back a yes for support but input doesn't come in the expected format.

nit: also can we rename the profile field to be `supportKittyKeyboardProtocol` or something a little more clear than just `kitty_keyboard`

cpendery:
> So I understand that we are no respecting the child requesting input to be sent via kitty keyboard; however, we are answering yes to child pty's with rio (& ghostty) that we support kitty keyboard. I think this could cause some weird failures in the future for apps that get back a yes for support but input doesn't come in the expected format.
> 
> nit: also can we rename the profile field to be `supportKittyKeyboardProtocol` or something a little more clear than just `kitty_keyboard`

Having a standard way of intercepting requests & replying with enable or disabled for features would be useful to extend for more responses as well, for example, we currently capture win32-input-mode (9001) requests and strip them. It would be great to centralize/standardize that logic

bindings/python/src/tui_test/_config.py
```diff
@@ -156,6 +156,8 @@ def normalize_profile(profile: object) -> Optional[Dict[str, Any]]:
     normalized = {}  # type: Dict[str, Any]
     if raw.get("scrollback") is not None:
         normalized["scrollback"] = raw["scrollback"]
+    if raw.get("kitty_keyboard") is not None:
+        normalized["kitty_keyboard"] = bool(raw["kitty_keyboard"])
```
cpendery, line 160: i'm pretty sure that this will convert invalid values to True by default, not sure if we want that behavior

aymanbagabas:
Yeah, I agree, maybe this one is doesn't make sense to have. A terminal emulator either supports Kitty Keyboard or doesn't and we shouldn't intercept the PTY/VT to test legacy handling. That should be part of the tui-test keyboard handling harness.

I've made it so that it tries to use the backend encoding capabilities and then fallback to our in-house one when a backend doesn't have key encoding capabilities in https://github.com/microsoft/tui-test/pull/182. I'll close this one and re-order the stack.

aymanbagabas:
Closing per the reasoning in https://github.com/microsoft/tui-test/pull/180#issuecomment-5593461393: a terminal either implements the Kitty keyboard protocol or it does not, and simulating the second case by gating the VT layer tests tui-test's own plumbing rather than a real terminal. That belongs in the keyboard harness, which #182 now does by routing to a backend's own encoder and falling back to the shared one.

The parts of this PR worth keeping have moved into #182:

- ghostty's missing `Emulator::keyboard_mode`, which left every key legacy-encoded on that backend whatever the child had negotiated
- `conformance_kitty_keyboard_modes_are_pushed_and_popped`, which pins push/pop for every backend that claims the protocol
- the declared `NoKittyKeyboard` divergence for xterm.js, tracked for the 6.1.0 bundle in #179

#182 is now based on `main` and the stack is reordered.

PR #181, opened 2026-08-26
feat(terminal): record OSC 8 hyperlinks on the cells they cover

aymanbagabas:
Adds OSC 8 hyperlinks to the cell vocabulary, so a test can assert where a link points rather than only what it says.

## Why

Nothing read a cell's hyperlink. `EmuCell` had no field for one, so every backend parsed `OSC 8` and then dropped it at the seam. A CLI that prints `View logs` as a link could be asserted on for its text and never for its URL, which is the part that breaks.

## The model

```rust
pub struct Hyperlink {
    pub id: Option<CompactString>,
    pub uri: CompactString,
}
// on EmuCell:
pub hyperlink: Option<Arc<Hyperlink>>,
```

Behind an `Arc` because a link covers a run of cells rather than one. Inline, a 40-character URL would be copied into every cell of the run and would more than double `EmuCell`; alacritty and rio already hand out reference-counted links, so this is also what they cost to clone.

Surfaced as `link` and `link_id` on the cell model, and as a `link` key in snapshots. The snapshot key records the URI rather than a boolean, on the same reasoning as the existing `underline` key recording a style rather than a boolean: a snapshot that only said "this is a link" would pass when the link silently started pointing somewhere else.

## Only `id` and the URI

OSC 8 allows arbitrary `key=value` params before the URI. All four backends keep `id` and discard the rest before a cell can be read back:

| backend | accessor | keeps |
| --- | --- | --- |
| alacritty | `Cell::hyperlink()` | `id`, `uri` |
| rio | `Extras::hyperlink` | `id`, `uri` |
| ghostty | `GridRef::hyperlink_uri()` | uri only |
| xterm.js | `oscLinkService.getLinkData(urlId)` | `id`, `uri` |

So there is nothing left to report, and reporting the full param list would mean parsing OSC 8 a second time ourselves in parallel with the emulator.

**Synthesized ids are dropped.** alacritty and rio each invent an id for a link that arrived without one, appending `_alacritty` or `_rio` to a process-wide counter. That is a rendering aid rather than something the child sent, it varies with the order links were parsed in, and no other backend produces it, so letting it through would put a backend's private counter into snapshots. A link with no `id=` reports `None`.

**ghostty reports no id at all**: its FFI exposes `ghostty_grid_ref_hyperlink_uri` and has no accessor for the parameter, so it never crosses the boundary. Declared as a `hyperlink_has_no_id` divergence rather than worked around.

## Two xterm.js bugs, both found by the new tests

**A link ran to the end of the row.** `getCell` reuses a single `CellData` instance and overwrites `extended` only for a cell that actually has extended attributes, so an unlinked cell read straight after a linked one still carried the previous cell's `urlId`. Now gated on `hasExtendedAttrs()`.

**A link forced a dashed underline.** `ExtendedAttrs.underlineStyle` returns `5` whenever a `urlId` is set:

```js
get underlineStyle(){return this._urlId?5:(469762048&this._ext)>>26}
```

That is how xterm.js chooses to draw links, not something the child sent, and the other three backends report what SGR said. Only reachable for a cell that is both underlined *and* linked, since a link alone never sets the underline flag. The raw bits still hold the real style, so those are read instead.

Both were latent before this PR: xterm.js has been reporting dashed underlines on underlined links since the backend landed, with no test that could see it.

## ghostty's viewport path

ghostty's render iterator exposes a cell's style and graphemes but not its hyperlink, so the URI has to come from a grid reference, which costs an FFI call per cell. `Row::has_hyperlink` narrows that to the rows that have one; it is documented as allowing false positives but not false negatives, so a row it skips genuinely has no link. A session with no links pays one flag check per row.

## Not configurable, by design

A hyperlink is something a cell carries, like its colours and attributes, so it is recorded the same way they are: always. There is no `hyperlinks` setting — nothing configures whether a cell reports its foreground either. (An earlier revision had one; it was dropped in review.)

## In the language bindings

`link` and `link_id` are on the JS `Cell` and the Python `Cell`, not just the Rust model, so a binding test can assert where a link points:

```
x=0 char='L'  link='https://example.com'  id='anchor'
x=3 char='K'  link='https://example.com'  id='anchor'
x=4 char=' '  link=''                     id=''
```

`link_id` carries an explicit `js_name` on the JS side to keep the snake_case its neighbours use, since napi would otherwise emit `linkId`.

## Tests

Five conformance cases, so all four backends are held to the same contract:

- `conformance_hyperlinks_cover_the_cells_they_open_over`
- `conformance_hyperlink_ids_are_only_what_the_child_sent`
- `conformance_adjacent_hyperlinks_stay_separate`
- `conformance_a_hyperlink_does_not_change_the_underline`
- `conformance_hyperlinks_can_be_turned_off`

`cargo test --workspace --features tui-test-rs/ghostty,tui-test-rs/rio,tui-test-rs/xtermjs` gives 435 passed, 0 failed.

Independent of #180; both touch `Profile` and the bindings, so they will need a trivial merge if both land.


README.md
```diff
@@ -301,6 +301,7 @@ Create `tui-test.toml`:
 ```toml
 [profiles.default]
 scrollback = 10000
+hyperlinks = true             # record OSC 8 links on the cells they cover
```
cpendery, line 304: maybe we can put all these features, like `kitty_keyboard` behind a subsection, like `terminal_features` or something like that

bindings/js/src/config.ts
```diff
@@ -115,6 +115,7 @@ function profileObject(value: unknown, name: string): Record<string, unknown> {
 
 export interface ProfilePayload {
```
cpendery, line 116: looks like we aren't including link or link_id in the language bindings, I think we should be exposing that

README.md
aymanbagabas: Dropped the config entirely rather than grouping it. A hyperlink is something a cell carries, like its colours and attributes, and nothing configures whether a cell reports its foreground either, so `hyperlinks` and its plumbing through `tui-test.toml` and both bindings are gone, along with the per-backend read gate and the test that only existed to check the disabled path.

The same reasoning retired the `kitty_keyboard` toggle you'd have grouped it with: #180 is closed, since a terminal either implements the protocol or it does not and faking the second case by gating the VT layer tests tui-test's own plumbing rather than a real terminal. What that PR carried worth keeping moved to #182.

So there is no `terminal_features` section to add — there are no terminal-feature settings left.

bindings/js/src/config.ts
aymanbagabas: Good catch — they stopped at the Rust boundary. `link` and `link_id` were on the core cell model and in `cells --json`, but neither binding carried them, so a JS or Python test could see a link's text and not where it pointed.

Both are now on the JS `Cell` (with `link_id` carrying an explicit `js_name` to keep the snake_case its neighbours use, since napi would otherwise emit `linkId`) and on the Python `Cell` dataclass.

Verified end to end rather than by inspection:

```
x=0 char='L'  link='https://example.com'  id='anchor'
x=1 char='I'  link='https://example.com'  id='anchor'
x=2 char='N'  link='https://example.com'  id='anchor'
x=3 char='K'  link='https://example.com'  id='anchor'
x=4 char=' '  link=''                     id=''
x=5 char='p'  link=''                     id=''
```

Note the file you commented on, `config.ts`, was the profile toggle rather than the cell shape; that toggle is gone now per the other thread.

crates/tui-test/src/terminal/ghostty/core.rs
```diff
@@ -54,7 +55,36 @@ fn underline(style: Underline) -> UnderlineStyle {
     }
 }
 
-fn cell_from_ghostty(cell: GhosttyCell, style: Style, graphemes: &[char]) -> Result<EmuCell> {
+/// Read a cell's OSC 8 URI off a grid reference.
+///
+/// Ghostty's FFI exposes the URI and nothing else, so the `id=` parameter is
+/// not recoverable here and every ghostty link reports `id: None`. That is
+/// declared as a conformance divergence rather than worked around.
+fn grid_hyperlink(grid: &GridRef<'_>) -> Result<Option<Arc<Hyperlink>>> {
```
cpendery: I'm pretty sure all these conversions are going to create lots of duplicate hyperlinks, can we use some sort of batch link table to reduce allocating the same URL N times when it is N cells long

crates/tui-test/src/assert/snapshot.rs
```diff
@@ -62,6 +62,15 @@ fn shift(prev: &EmuCell, cur: &EmuCell) -> Map<String, Value> {
     if prev.underline != cur.underline {
         m.insert("underline".into(), json!(cur.underline.name()));
     }
+    // The URI, not a boolean: a snapshot that only recorded "this is a link"
+    // would pass when the link silently started pointing somewhere else,
+    // which is the whole thing worth asserting about a link. The `id=` is
+    // deliberately not recorded, because it is an anchor for joining a
+    // wrapped link rather than anything a reader can check, and the ghostty
+    // backend cannot report it at all.
+    if prev.uri() != cur.uri() {
```
cpendery, line 71: maybe we can define a new property like `include_metadata` instead of using `include_colors` to indicate if a url should be included. I think a URL is less of a style and more of a property. Might make sense to refine the other to `include_style` instead of `include_colors` since we use getByStyle() locators

bindings/js/native/index.d.ts
```diff
@@ -91,6 +91,10 @@ export interface Cell {
   underline: boolean
   underline_style: UnderlineStyle
   underline_color: Color
+  /** The OSC 8 URI this cell links to, empty when it links nowhere. */
+  link: string
+  /** The link's `id=` parameter, empty when the sequence carried none. */
+  link_id: string
```
cpendery, line 105: let's make sure we keep this private, I don't think we want to expose this to end users, just the link

aymanbagabas: that was intentional. sometimes you want to assert wrapped links having the same program id when a program handles wrapping.

crates/tui-test/src/terminal/ghostty/core.rs
aymanbagabas: Done in 1ce6068 — both capture paths now share a per-capture table keyed by URI and hand out clones of one `Arc`, so an N-cell link allocates the URI once instead of N times. It also dedupes a link that wraps across rows, which the old code re-read per row. The xterm.js shim already batched links this way, so the two backends now agree.

Good catch — the scrollback path (`full_rows`) was the worse of the two, since it hits every cell rather than only the rows `has_hyperlink` flags.

crates/tui-test/src/assert/snapshot.rs
aymanbagabas: Renamed to `include_style` in 6dbc112. Agreed the old name was wrong independently of links: it already recorded bold, italic, the underline shape and the underline color, so naming it after colors made everything else look undocumented.

I went with one flag rather than adding `include_metadata` beside it. I went looking for how emulators actually bucket OSC 8 before deciding:

| | where the link lives | in the SGR reset group? |
|---|---|---|
| alacritty | `CellExtra { zerowidth, underline_color, hyperlink }` | no |
| xterm.js | `ExtendedAttrs { underlineStyle, underlineColor, urlId }` | no |
| VTE | `VteCellAttr { attr, m_colors, hyperlink_idx }` | no — `reset_sgr_attributes()` touches `attr` and `m_colors` only |
| ghostty | separate `hyperlink_map`/`hyperlink_set`, beside the style set | no |
| kitty | `active_hyperlink_id`, outside `Cursor.sgr` | no |

So you're right that it isn't a style, and every one of them agrees: `SGR 0` clears every attribute and leaves the link running. But they also all store it *with* the cell's appearance, because it behaves the same way — set on the cursor, inherited by every cell written while it's open.

I added that as a conformance test (808e43f), and all four of our backends agree.

That's why it's one switch: a reader who asks for style and gets a box back with no links has been told something false about the frame, and a second flag is one people forget. The doc comment now says a link isn't an SGR attribute so the name doesn't overclaim.

README.md
aymanbagabas: Moot now — there is no config to group. `kitty_keyboard` went away with #180 (a terminal either speaks the protocol or it does not, so there was nothing for a user to choose), and `hyperlinks` went away here after your other comment, since a link is recorded like any other cell property rather than opted into. This PR adds no config keys at all.

bindings/js/native/index.d.ts
aymanbagabas: Kept it, but documented rather than silent — I want to make the case, since this crosses with your earlier comment asking for both to be exposed.

`id=` is the one thing on a cell that nothing else can express: it is what joins the runs of *one* logical link. A program that wraps its own links tags each run with a shared id so a terminal can highlight them together, and without the field there is no way to assert that two runs are the same link rather than two links that happen to point at the same place.

Your concern is real though, and it is narrower than the field being useless: `link_id` is the **only** backend-dependent field on a cell. Ghostty exposes `ghostty_grid_ref_hyperlink_uri` and nothing else, so it is always empty there, and "no `id=` was sent" cannot be told apart from "this backend cannot see it". That is now stated on the field in all three bindings rather than left for someone to discover, and ghostty-org/ghostty#14034 is open upstream to close the gap.

Where I did follow you: **no locator matches on it.** `--link` filters on the URI only, and there is deliberately no `--link-id`. A read-only field that a test can print and reason about is a different risk from an assertion that quietly means something else after a backend switch.

bindings/js/src/config.ts
aymanbagabas: Done — `link` and `link_id` are on the `Cell` type in both the JS and Python bindings now, and in `cells --json`. This was the real finding in that review: the data stopped at the Rust boundary, so nothing outside the crate could see it.

Follow-up on `link_id` specifically in the newer thread on `index.d.ts`, where you asked to keep it private.

crates/tui-test/src/terminal/ghostty/core.rs
cpendery: This was an issue for rio & alacritty as well, but the fix only covered ghostty

aymanbagabas: You're right, and they were the worse case. Fixed in b47861d.

I measured before and after — 80x24 fully covered by one link, against the same grid with no link, release build:

| backend | before | after | vs plain grid |
|---|---|---|---|
| alacritty | 105µs | **18µs** | 13.7x → 2.3x |
| rio | 100µs | **19µs** | 13.6x → 2.4x |
| ghostty | 26µs | 23µs | 1.6x → 1.5x |
| xterm.js | unchanged | | already batched |

So alacritty and rio were paying ~13x for a linked grid while the path I'd fixed paid 1.6x. xterm.js turned out to be fine already: `decode_links` builds the table once per batch and cells clone the `Arc`.

The fix also replaced the map I gave ghostty. A run is contiguous — that's what makes it a run — so a one-entry cache hits for every cell after the first, including across a wrap where the next row continues the same link. It's less code than a map and costs no hash per cell. It lives in `cell.rs` now, so all three converters share one mechanism instead of three.

One correctness note while doing it: the cache compares the `id=` as well as the URI. Keying on the URI alone would have let two links pointing at the same place under different ids collapse into one, and the second run would have inherited the first one's id. There are now conformance tests for both — that a run shares one `Arc` (pointer equality), and that same-URI-different-id stays distinct.

cpendery (approved):

PR #182, opened 2026-08-26
feat(input): route key encoding to the backend, honor DECCKM, and report ghostty's keyboard modes

aymanbagabas:
Routes key encoding to a backend's own encoder where one exists, fixes a bug that finding out about surfaced, and implements a `keyboard_mode` that ghostty never had.

## Three problems

**`keys.rs` modelled no DECCKM at all.** `CSI ?1h` moves the cursor keys from `CSI A` onto `SS3 A`, and readline, vim and less all set it — so `key press Up` has been sending `CSI A` to children that asked for `SS3 A`.

**ghostty never reported any Kitty flags.** `Emulator::keyboard_mode` has a default returning `KeyboardMode::empty()`, and `GhosttyEmu` never overrode it, so every key reaching a ghostty session was legacy-encoded no matter what the child had negotiated. `libghostty-vt` exposed the state as `Terminal::kitty_keyboard_flags()` the whole time; it was simply never called.

**Encoding ignored terminal state the backends already track.** Keypad application mode, `modifyOtherKeys`, the alt-escape prefix — none of it was modelled.

## The routing

`Emulator::encode_key(&KeyPress) -> Option<Vec<u8>>`, defaulting to `None`. A backend with an encoder uses it; `None` falls back to the shared one.

| backend | own encoder? | why |
| --- | --- | --- |
| ghostty | **yes** | `ghostty_vt::key::Encoder` |
| alacritty | no | encoding lives in the `alacritty` GUI crate, not `alacritty_terminal` |
| rio | no | same split; `rio-vt` has only hex and graphics encoders |
| xtermjs | no | `evaluateKeyboardEvent` is in `browser/`, absent from the headless bundle |

Ghostty's `set_options_from_terminal` reads the live terminal, so routing to it applies the modes the shared encoder does not model.

**Declining is a feature.** Ghostty's modifier bitmask has no Hyper or Meta, which the Kitty protocol does define. Rather than encode without them, `encode_key` returns `None` and the event falls back, so a modifier is never silently dropped. Fallback is all-or-nothing per token, since a token half-encoded by each would interleave two encodings of one keypress.

## DECCKM

Reaches exactly the keys `CSI ... A`-`D`, `H` and `F` spell, and only unmodified presses — which is already what `ss3_unmodified` means for F1-F4, so it reuses that rather than adding a second rule that could drift. A modified cursor key stays `CSI` (`SS3` has nowhere to put the parameter), and the Kitty protocol outranks the mode entirely.

Ghostty encodes its own DECCKM; the shared encoder's covers the other three. Without both halves this PR would have made ghostty correct and left the others wrong — a divergence introduced by a fix.

## Four bugs found by #183

The differential test in #183 was written against this branch and immediately found four bugs in it:

| symptom | cause |
| --- | --- |
| `key press A` sent `a` | tokens were lowercased wholesale — right for a named key, wrong for a character |
| `key press Space` sent nothing | ghostty encodes a text-bearing key with no `utf8` as empty, and only single-character tokens set it |
| `key press Alt+a` sent `a` | `set_options_from_terminal` resets `macos_option_as_alt` to `False`, so Alt composed instead of prefixing ESC |
| `Shift+a` sent `CSI 97;2u` instead of `A` | `consumed_mods` was never set, so ghostty reported a Shift already spent producing the text |

`KeyPress` now carries the *unshifted* key plus the `text` it produces, so `!` arrives as `"1"` with shift and text `"!"`. Text is carried rather than derived because deriving it needs a keyboard layout.

## Not in this PR any more

The `kitty_keyboard` profile toggle that #180 proposed. A terminal either implements the protocol or it does not, and simulating the second case by gating the VT layer tests tui-test's own plumbing rather than a real terminal — that belongs in the keyboard harness, which is what this PR builds. #180 is closed; what it carried that was worth keeping lives here: ghostty's `keyboard_mode`, the push/pop conformance case, and the `NoKittyKeyboard` divergence for xterm.js (tracked for the 6.1.0 bundle in #179).

## Shape

`InputModes` groups the state encoding branches on, and public entry points take `impl Into<InputModes>` with `From<KeyboardMode>`, so existing callers and all 32 existing test call sites compile untouched. `Emulator::cursor_key_application` is implemented by all four backends.

## Tests

Five DECCKM cases, four routing cases against a real ghostty terminal, and a conformance case pinning Kitty push/pop for every backend that claims the protocol — the case that would have caught the ghostty gap.

Based on `main`; #183 stacks on this.

### DIFF
--- a/crates/tui-test/src/terminal/ghostty/core.rs
+++ b/crates/tui-test/src/terminal/ghostty/core.rs
@@ -320,15 +320,33 @@ impl GhosttyCore {
             event.set_consumed_mods(GhosttyMods::SHIFT);
         }
 
-        // A key that produces text has to say so. Ghostty encodes nothing at
-        // all for a text-bearing key with no `utf8` set, and needs it for the
-        // associated-text and alternate-key parts of the Kitty protocol.
-        // `unshifted_codepoint` is what the key sits on, which is exactly what
-        // `KeyPress::key` already holds for a character.
-        let mut chars = press.key.chars();
-        if let (Some(ch), None) = (chars.next(), chars.next()) {
+        // The codepoint the key sits on. Ghostty builds the Kitty `CSI <code> u`
+        // form from it, so a key without one is encoded as its text alone and
+        // silently loses its modifiers: `Ctrl+Space` came out as a plain space.
+        //
+        // For a character the key *is* the codepoint. For a named key it is
+        // not — `space` is a word — so the text it produces stands in, which
+        // is how `space` gets U+0020. Named keys that produce no text, such as
+        // Enter and Tab, need nothing here: ghostty knows their codepoints
+        // from the `Key` itself.
+        //
+        // Taken from the key before the text so that `Shift+a` reports `a`
+        // rather than the `A` it produced.
+        let single = |s: &str| {
+            let mut chars = s.chars();
+            match (chars.next(), chars.next()) {
+                (Some(ch), None) => Some(ch),
+                _ => None,
+            }
+        };
+        if let Some(ch) = single(&press.key).or_else(|| press.text.as_deref().and_then(single)) {
             event.set_unshifted_codepoint(ch);
         }
+
+        // A key that produces text has to say so: ghostty encodes nothing at
+        // all for a text-bearing key with no `utf8` set, and needs it for the
+        // associated-text and alternate-key parts of the Kitty protocol. A
+        // release produces no text, only the key that was let go.
         if press.event != KeyEventKind::Release {
             if let Some(text) = press.text.as_deref() {
                 event.set_utf8(Some(text));
--- a/crates/tui-test/src/terminal/ghostty/core.rs
+++ b/crates/tui-test/src/terminal/ghostty/core.rs
@@ -345,12 +345,15 @@ impl GhosttyCore {
 
         // A key that produces text has to say so: ghostty encodes nothing at
         // all for a text-bearing key with no `utf8` set, and needs it for the
-        // associated-text and alternate-key parts of the Kitty protocol. A
-        // release produces no text, only the key that was let go.
-        if press.event != KeyEventKind::Release {
-            if let Some(text) = press.text.as_deref() {
-                event.set_utf8(Some(text));
-            }
+        // associated-text and alternate-key parts of the Kitty protocol.
+        //
+        // This is the text the key produces on the current layout rather than
+        // anything the event does with it, so it is set for a release too:
+        // ghostty derives the shifted alternate key from it, and withholding
+        // it made `Shift+a` release as `CSI 97;2:3u` instead of the
+        // `CSI 97:65;2:3u` its press had already reported.
+        if let Some(text) = press.text.as_deref() {
+            event.set_utf8(Some(text));
         }
 
         let mut encoder = Encoder::new().context("creating key encoder")?;
--- a/crates/tui-test/src/terminal/ghostty/core.rs
+++ b/crates/tui-test/src/terminal/ghostty/core.rs
@@ -9,6 +9,10 @@ use base64::engine::general_purpose::STANDARD as BASE64;
 use base64::Engine as _;
 use compact_str::CompactString;
 use ghostty_vt::error::Error as GhosttyError;
+use ghostty_vt::key::{
+    Action as GhosttyKeyAction, Encoder, Event as GhosttyKeyEvent, Key, KittyKeyFlags,
+    Mods as GhosttyMods, OptionAsAlt,
+};
 use ghostty_vt::render::{CellIterator, CursorVisualStyle, RowIterator};
 use ghostty_vt::screen::{Cell as GhosttyCell, CellContentTag, CellWide, GridRef};
 use ghostty_vt::style::{Palette, PaletteIndex, RgbColor, Style, StyleColor, Underline};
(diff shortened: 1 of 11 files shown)


crates/tui-test/src/terminal/ghostty/core.rs
```diff
@@ -238,6 +243,114 @@ impl GhosttyCore {
...
+            event.set_consumed_mods(GhosttyMods::SHIFT);
+        }
+
+        // A key that produces text has to say so. Ghostty encodes nothing at
+        // all for a text-bearing key with no `utf8` set, and needs it for the
+        // associated-text and alternate-key parts of the Kitty protocol.
+        // `unshifted_codepoint` is what the key sits on, which is exactly what
+        // `KeyPress::key` already holds for a character.
+        let mut chars = press.key.chars();
+        if let (Some(ch), None) = (chars.next(), chars.next()) {
```
cpendery: Seems like this will only work for keys with single character names. Space will get translated to "space" so U+0020 doesn't get properly emitted. `ctrl+space`, `alt+space`, `space (key-up)` won't get sent properly

aymanbagabas: You're right, and it was worse than not emitting U+0020 — in Kitty mode the modifier vanished entirely.

`unshifted_codepoint` was only set for single-character key names, so `space` never got one. Ghostty builds the `CSI <code> u` form from that codepoint, and with none it fell back to encoding the key as its bare text:

| mode | `ctrl+space` | `space` key-up |
|---|---|---|
| legacy | `\0` ✓ | `""` ✓ |
| kitty disambiguate | `" "` ✗ | — |
| kitty report-all | `" "` ✗ | `""` ✗ |

Legacy hid it because there the text *is* the answer and ghostty derives Ctrl from the mods itself.

Fixed in 64ed2d2 by taking the codepoint from the key when the key is a character and otherwise from the text it produces, which is how `space` gets U+0020. Reading the key first keeps `shift+a` reporting `a` rather than the `A` it produced. Enter and Tab need nothing — ghostty knows their codepoints from the `Key`.

Pulling that thread turned up two more, both the same shape of withholding something from the encoder:

- c0523ff — `utf8` was also gated on the action, but ghostty documents it as "the text generated by the key for the current keyboard layout" and derives the shifted alternate key from it. `shift+a` pressed as `CSI 97:65;2u` and released as `CSI 97;2:3u`, reporting an alternate key on the way down and forgetting it on the way up. It does not leak text into a release: ghostty already withholds associated text there.
- c99cfec — the shared encoder sent nothing at all for the release of a text-producing key unless report-all was on. The spec says event-types reporting "causes the terminal to report key repeat and key release events", with a note exempting exactly Enter, Tab and Backspace "so that the user can still type reset at a shell prompt"; naming three exemptions only makes sense if every other key is reported. kitty's `key_encoding.c:398` agrees. Repeats are left alone — holding a key down means more text.

All three now agree with ghostty on every event in the matrix.

cpendery:
two more small fixes
1) Shift+Space is only going to send " " instead of `CSI 32;2 u`
2) we use ` has_kitty_only_modifier` to reject Super for Ghostty, but it does support super. Maybe we can change this based on the emulator

aymanbagabas:
Two good catches, and they went different ways — thanks for both.

## 2) Super — you're right, fixed in 67502cc

Ghostty's bitmask is Ctrl, Alt, Shift and **Super**; only Hyper and Meta are missing, and `ghostty_mods` already mapped Super. The guard reused the shared encoder's `has_kitty_only_modifier`, which groups all three — but that predicate answers *"has no legacy form"*, not *"ghostty cannot carry it"*. Two different questions. The doc comment and the existing test both already said hyper/meta only, so the code was the odd one out.

There was a real half to it though. In **legacy** mode ghostty drops Super silently:

| | before | after |
|---|---|---|
| `super+a` (legacy) | `""` — nothing at all | declines → shared encoder |
| `ctrl+super+a` (legacy) | `CSI 97;5u` — that's `ctrl+a`, Super gone | declines → shared encoder |
| `super+a` (any Kitty mode) | declined | `CSI 97;9u` ✓ |
| `ctrl+super+a` (any Kitty mode) | declined | `CSI 97;13u` ✓ |

So it now declines Super only where the legacy encoding would eat it, and hands it over everywhere else, matching the shared encoder exactly.

## 1) Shift+Space — I think this one is correct as-is

I checked kitty's source rather than reasoning about it. `encode_glfw_key_event` has:

```c
bool send_text_standalone = !ev.report_text;
...
if (send_text_standalone && ev.has_text && (ev.action == PRESS || ev.action == REPEAT)) return SEND_TEXT_TO_CHILD;
```

`report_text` is flag **8** (report-all-keys-as-escape-codes). So until that flag is set, *any* press carrying text is sent as text, modified or not — it never reaches `encode_key`. `Shift+Space` carries " ", so a space is what kitty sends. `encode_printable_ascii_key_legacy` agrees independently: `if (ev->mods.value == SHIFT) return ... "%c"`.

Measured across the modes, ghostty and the shared encoder agree everywhere:

| flags | Down | Up |
|---|---|---|
| legacy | `" "` | `""` |
| `>1u` disambiguate | `" "` | `""` |
| `>3u` +events | `" "` | `CSI 32;2:3u` |
| `>15u` all | `CSI 32;2u` | `CSI 32;2:3u` |

So `CSI 32;2u` does happen — once keys are reported as escape codes, which is the flag that asks for it.

**But you found a real gap:** `Shift+Space` had no test anywhere, and neither did `Space` or any Super chord in the oracle. All are now covered — the four rows above are pinned as a test, and `Space`/`Shift+Space`/`Ctrl+Space`/`Alt+Space`/`Super+a`/`Ctrl+Super+a` are in the oracle's case list. Still 0 unexplained disagreements.

If you were thinking of a specific terminal that does send `CSI 32;2u` under plain disambiguation, say which and I'll go look — I may be reading the flag boundary wrong.

cpendery (dismissed):

cpendery (approved):

PR #183, opened 2026-08-26
test(input): hold the shared key encoder against ghostty's

aymanbagabas:
Differential-tests tui-test's shared key encoder against ghostty's, which is a reference implementation maintained by people who work on nothing else.

> **Stacked on #182**, which is now based on `main`. The oracle needs both `encode_key` and ghostty's `keyboard_mode` from that PR — without the latter it compared ghostty in Kitty mode against the shared encoder in legacy mode, producing 329 phantom disagreements. Review #182 first; this PR's own change is a single test file.

## Why

Three of the four backends have no key encoder of their own and always fall back to `keys.rs`. A bug there is a bug on alacritty, rio, and xterm.js at once — and nothing was checking `keys.rs` against anything but its own expectations. Ghostty ships an encoder, so it can be held up as an oracle.

## It has already paid for itself

Four real bugs, all in the PRs beneath this one, all found by this test and none by the hand-written cases:

| symptom | cause |
| --- | --- |
| `key press A` sent `a` | tokens were lowercased wholesale, right for a named key and wrong for a character |
| `key press Space` sent nothing | ghostty encodes a text-bearing key with no `utf8` as empty, and only single-character tokens set it |
| `key press Alt+a` sent `a` | `set_options_from_terminal` resets `macos_option_as_alt` to `False`, so Alt composed instead of prefixing ESC |
| `Shift+a` sent `CSI 97;2u` instead of `A` | `consumed_mods` was never set, so ghostty reported a Shift that had been spent producing the text |

All four would have shipped in #182. That is the argument for the test.

## What it compares

495 events: every token whose encoding both encoders claim to define, across five terminal states — legacy, application cursor keys (`CSI ?1h`), and three Kitty mode combinations (`>1u`, `>3u`, `>15u`).

## Two spellings normalized

Both are spellings, not meanings, and the spec makes both legal:

- **Event type.** With `REPORT_EVENT_TYPES` on, ghostty writes the press event as `:1`; kitty and xterm omit it because press is the default. `CSI 1;5:1A` and `CSI 1;5A` decode identically.
- **Default parameters.** Ghostty writes `CSI 1;1A` where kitty and xterm write `CSI A`.

Normalizing keeps those modes in the matrix rather than excluding them. Both normalizers are deliberately narrow — the event-type one only strips a `:1` immediately before the final byte, so the `:65` alternate-key sub-parameter in `CSI 97:65;6u` is untouched.

## What is excluded, and why

Combinations the legacy scheme cannot express. Ghostty upgrades them to `CSI-u` or xterm's `modifyOtherKeys` form: `Ctrl+Tab` becomes `CSI 27;5;9~` where legacy has a bare `\t` with the Ctrl silently lost, and `Ctrl+Shift+0` becomes `CSI 41;5u`. That is a richer answer to a different question, not a different answer to this one, so asserting it and then explaining it away would only teach readers to ignore the test. F3 is excluded too: ghostty and kitty spell it `CSI 13~`, xterm spells it `SS3 R`, and that split predates both.

## The one real disagreement

Named in `ORACLE_KNOWN_DIVERGENCES` with its reasoning rather than deleted:

> Under report-all-keys, ghostty sends Space as the text `` ` ` `` where the shared encoder sends `CSI 32u`. Report-all-keys means what it says — the Kitty spec has text-producing keys report as escape codes under this flag — so the shared encoder looks right here and it is worth raising upstream.

The test asserts that a listed divergence **still diverges**, so the list cannot silently go stale: if ghostty changes, the entry fails as loudly as a new disagreement would.

## Result

495 compared, 0 unexplained disagreements, 1 documented. `cargo test --workspace --features tui-test-rs/ghostty,tui-test-rs/rio,tui-test-rs/xtermjs` gives 533 passed, 0 failed.

### DIFF
--- a/crates/tui-test/src/terminal/emu.rs
+++ b/crates/tui-test/src/terminal/emu.rs
@@ -263,7 +263,6 @@ pub trait Emulator: Send {
     fn bracketed_paste_mode(&self) -> bool {
         false
     }
-
     /// Encode one key event with the backend's own key encoder.
     ///
     /// `None` means the backend has no encoder, or has one that cannot express
--- a/crates/tui-test/src/terminal/ghostty/mod.rs
+++ b/crates/tui-test/src/terminal/ghostty/mod.rs
@@ -719,32 +719,80 @@ mod tests {
     /// reporting is on; kitty and xterm omit them. As with the event type,
     /// this normalizes a spelling rather than a meaning.
     fn normalize_default_params(bytes: &[u8]) -> Vec<u8> {
-        let Some(rest) = bytes.strip_prefix(b"\x1b[1;1") else {
-            return bytes.to_vec();
-        };
-        // Only when what remains is the final byte, so a longer parameter
-        // list that merely starts `1;1` is left alone.
-        if rest.len() == 1 && (rest[0].is_ascii_alphabetic() || rest[0] == b'~') {
-            let mut out = b"\x1b[".to_vec();
-            out.extend_from_slice(rest);
-            return out;
+        const DEFAULT_PARAMS: &[u8] = b"\x1b[1;1";
+        let mut out = Vec::with_capacity(bytes.len());
+        let mut i = 0;
+        while i < bytes.len() {
+            // Rewritten wherever it appears rather than only at the start,
+            // since one action can encode as more than one sequence.
+            //
+            // Only when the final byte follows immediately, so a longer
+            // parameter list that merely starts `1;1` is left alone.
+            let is_default_params = bytes[i..].starts_with(DEFAULT_PARAMS)
+                && bytes
+                    .get(i + DEFAULT_PARAMS.len())
+                    .is_some_and(|b| b.is_ascii_alphabetic() || *b == b'~');
+            if is_default_params {
+                out.extend_from_slice(b"\x1b[");
+                i += DEFAULT_PARAMS.len();
+                continue;
+            }
+            out.push(bytes[i]);
+            i += 1;
         }
-        bytes.to_vec()
+        out
     }
 
     /// Events where the two encoders genuinely disagree, with the reason.
     ///
     /// Kept as a named list rather than dropped from the matrix, so the
-    /// disagreement stays visible and a change to either side still shows up
-    /// as this list going stale.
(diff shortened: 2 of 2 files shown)


cpendery (commented):
Let's cover a case of sending a keypress event as ghostty will send two events for this, while the shared encoder will send 1

aymanbagabas:
> Let's cover a case of sending a keypress event as ghostty will send two events for this, while the shared encoder will send 1

Added in 6a80c30 — the oracle now runs every action, and `Press` earns its place immediately, because it is the only one that puts two events side by side.

It found a bug in the oracle itself: `normalize_default_params` only rewrote a buffer holding exactly one sequence, so as soon as a down and its release were concatenated the normalization stopped applying and **every functional key looked like a disagreement** — `Up` as `CSI 1;1A CSI 1;1:3A` against `CSI A CSI 1;1:3A`. It now rewrites the default parameter pair wherever it appears.

While widening it I also added `Repeat`, which pinned down a distinction worth having: a repeat of a text-producing key stays text, because holding a key down means more text. Only a key with no text of its own repeats as an escape code.

That took the matrix from 495 events to **1980** across five terminal states, and shook out three real bugs now fixed down in #182 (64ed2d2, c0523ff, c99cfec).

One genuine disagreement is left, and I've kept the named-list mechanism for it rather than dropping it from the matrix, so it stays visible and a change on either side shows up as the list going stale. Enter, Tab and Backspace report no release until report-all is set; kitty gates that exemption on the chord (`key_encoding.c` wraps both exemption blocks in `if (!ev->mods.value)`) while ghostty gates it on the key (`key_encode.zig` switches on `event.key` with no modifier test), so they differ on a modified one such as `shift+tab`. The shared encoder follows kitty, whose author wrote the spec; the spec's own wording ("the Enter, Tab and Backspace keys") is short of deciding it. Worth raising upstream.

Worth noting the list previously held `Space` as a divergence to raise upstream. That turned out to be our own bug, which is exactly what widening the comparison exposed — so the entries above are now backed by reading both implementations rather than by one side disagreeing.

cpendery (dismissed):

cpendery (dismissed):

cpendery (dismissed):

cpendery (approved):

PR #203, opened 2026-09-03
feat(assert): match a cell by its OSC 8 link

aymanbagabas:
Makes OSC 8 links assertable, not just readable.

> **Stacked on #181.** That PR records links on cells; this one adds the assertion surface. Review #181 first.

## Why

#181 makes links observable through `cells --json` and snapshots, but the style query has no link field. So a test that wants to check *where* a link points has to shell out and parse JSON:

```sh
tui-test cells 0 0 4 1 --json | jq -r '.data.cells[0].link'
```

That is precisely what the style query exists to avoid. `--fg`, `--underline-color` and the rest are all first-class; the link was not.

## What

`link` joins the other style properties, so it works everywhere they already do:

```sh
tui-test expect text "Docs" --link "https://example.com"
tui-test expect text "plain" --link ""
```

```ts
await term.getByStyle({ link: "https://example.com" }).expect();
```

```python
await term.get_by_style(TextStyle(link="https://example.com")).expect()
```

Because it is an ordinary style property it also composes with the `within`, `after` and `before` locator stages for free.

## An empty string is a requirement, not an absence

`--link ""` asks for a cell that links **nowhere**. That is how a test asserts `OSC 8 ;;` actually closed a link, rather than only that some *other* link is absent — a distinction that matters because a link running to the end of the row is exactly the bug #181 found in the xterm.js shim.

The `Option` already carries "the caller did not ask", which leaves the inner `String` free to mean something. Hence `Option<String>` rather than a bare `String` with a sentinel.

## Verification

Unit tests cover the matcher: matching URL, non-matching URL, empty-link-vs-linked-cell in both directions, and a style with no link still matching a linked cell (so this does not silently narrow existing queries). A CLI test pins that `--link ""` survives parsing as an empty string rather than collapsing to `None`.

Checked against a live session too, since a parse test would not catch a wiring mistake:

| assertion | exit |
| --- | --- |
| `expect text "LINK" --link "https://example.com"` | 0 |
| `expect text "LINK" --link "https://wrong.example"` | 1 |
| `expect text "plain" --link ""` | 0 |
| `expect text "plain" --link "https://example.com"` | 1 |
| `expect text "LINK" --link ""` | 1 |

`cargo test --workspace --features tui-test-rs/ghostty,tui-test-rs/rio,tui-test-rs/xtermjs` gives 633 passed, 0 failed. `cargo fmt --check` and both `clippy -D warnings` invocations are clean.

### DIFF
--- a/bindings/js/native/index.d.ts
+++ b/bindings/js/native/index.d.ts
@@ -165,6 +165,7 @@ export interface LocatorStyle {
   hidden?: boolean
   strikethrough?: boolean
   blink?: boolean
+  link?: string
 }
 
 export interface MouseClickOptions {
--- a/bindings/js/native/lib.rs
+++ b/bindings/js/native/lib.rs
@@ -461,6 +461,7 @@ pub struct LocatorStyle {
     pub hidden: Option<bool>,
     pub strikethrough: Option<bool>,
     pub blink: Option<bool>,
+    pub link: Option<String>,
 }
 
 #[napi(object)]
@@ -647,6 +648,7 @@ fn core_style(style: LocatorStyle) -> CoreTextStyle {
         hidden: style.hidden,
         strikethrough: style.strikethrough,
         blink: style.blink,
+        link: style.link,
     }
 }
 
--- a/bindings/js/src/client.ts
+++ b/bindings/js/src/client.ts
@@ -79,6 +79,8 @@ export interface TextStyleExpectation {
   hidden?: boolean;
   strikethrough?: boolean;
   blink?: boolean;
+  /** Required OSC 8 link target. Pass `""` to require no link. */
+  link?: string;
 }
 
 export interface LocatorWaitOptions {
@@ -320,6 +322,7 @@ function textStyleValue(style: TextStyleExpectation): RuntimeLocatorStyle {
     hidden: style.hidden,
     strikethrough: style.strikethrough,
     blink: style.blink,
+    link: style.link,
   };
 }
 
--- a/bindings/python/native/src/lib.rs
+++ b/bindings/python/native/src/lib.rs
@@ -1457,6 +1457,7 @@ fn core_style(dict: &Bound<'_, PyDict>) -> Result<TextStyle, TuiTestError> {
         hidden: py_bool(dict, "hidden")?,
         strikethrough: py_bool(dict, "strikethrough")?,
         blink: py_bool(dict, "blink")?,
+        link: py_string(dict, "link")?,
     })
 }
 
--- a/bindings/python/src/tui_test/types.py
+++ b/bindings/python/src/tui_test/types.py
@@ -111,6 +111,7 @@ class TextStyle:
     hidden: Optional[bool] = None
     strikethrough: Optional[bool] = None
     blink: Optional[bool] = None
+    link: Optional[str] = None
 
 
 @dataclass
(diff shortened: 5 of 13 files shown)


cpendery (commented):
I think it makes more sense to have a `getByUrl` or `getByMetadata` locator instead of bundling it in with getByStyle.

Also consider `A B` where the space as a url on it, asserting that `A B` has no link will pass

aymanbagabas:
> Also consider `A B` where the space has a url on it, asserting that `A B` has no link will pass

Confirmed, and it was worse than that in both directions. Fixed in 840b811 with a regression test that reproduces exactly this case.

The cause was pre-existing: `located_match_has_style` skips blank cells when a match has any visible cell. That is right for a color — requiring the foreground of a space would fail a match for a reason no reader could see — but wrong for a link, which is a region of the screen rather than an appearance, and which a space genuinely carries. So `A B` whose space alone linked somewhere satisfied `--link ""`, and `A B` whose space alone linked *nowhere* satisfied `--link https://…`. Both now fail; a run matches only when every cell agrees, blanks included.

> I think it makes more sense to have a `getByUrl` or `getByMetadata` locator instead of bundling it in with getByStyle.

I looked at how emulators bucket OSC 8 before deciding, and the answer was consistent: it is stored *with* the cell's appearance everywhere, but is a member of the SGR group nowhere.

| | where the link lives | in the SGR reset group? |
|---|---|---|
| alacritty | `CellExtra { zerowidth, underline_color, hyperlink }` | no |
| xterm.js | `ExtendedAttrs { underlineStyle, underlineColor, urlId }` | no |
| VTE | `VteCellAttr { attr, m_colors, hyperlink_idx }` | no — `reset_sgr_attributes()` skips it |
| ghostty | separate `hyperlink_map`/`hyperlink_set` | no |
| kitty | `active_hyperlink_id`, outside `Cursor.sgr` | no |

I pinned the consequence as a conformance test on #181: `SGR 0` does not close a link on any of our four backends.

So you're right that it isn't a style, and I've said so in the doc comment rather than letting the name overclaim. I kept it in the style query anyway because the query is a conjunction over cell properties, and a separate locator cannot express "the bold cell that links here" — `within` means *inside*, not *and*, so it would have to be intersected by hand.

On `id=`: it stays readable on a cell but deliberately unmatchable. It is worth asserting on in principle — it is what joins the runs of one logical link — but ghostty's FFI cannot report it at all, so a query against it would quietly mean different things on different backends.

cpendery:
What about `A B` in a single link, bolded text, and unbolded space. It will match bold and URL separately, but fails when combined. we might need to preserve the existing whitespace policy for visual styles.

aymanbagabas:
Right, and that was a real regression in my fix. Fixed in the latest push.

My change made the whitespace policy a property of the **match** — "if a link was asked for, check every cell" — when it should be a property of the **field**. So the whole style got applied to the blank, and your case fell exactly in the gap:

`A B`, linked throughout, letters bold, space not (the usual shape — nothing bolds a space):

| query | before | after |
|---|---|---|
| `--bold` | matches | matches |
| `--link ...` | matches | matches |
| `--bold --link ...` | **fails** | matches |

The two halves answer different questions, so they now get asked separately:

- **an appearance** skips a blank, which cannot show it — the existing policy, preserved as you suggested
- **a link** is a region of the screen rather than an appearance, and a space inside one is part of the link, so it is checked on every cell

Splitting them is what makes the combination mean what each half means. The earlier `A B`-with-a-linked-space case still fails as it should, since that is the link half.

There is a regression test for your exact scenario now, and I verified it fails against the version you were reading — the assertion that trips is "so asking for both matches too".

cpendery (dismissed):

cpendery (approved):

PR #208, opened 2026-09-05
feat: add Go support

meza:
Adds Go bindings for terminal automation using the same in-process Rust engine as the JavaScript and Python bindings.

This PR contains the public Go API, its purego runtime bridge, and the Rust native adapter. Source checkouts build the adapter and select it explicitly with TUI_TEST_GO_NATIVE_LIBRARY.

- Provides typed clients, input controls, inspection, locators, waits, assertions, screenshots, and recording.
- Adds testing.TB helpers for session cleanup and failure artifacts.
- Keeps required runtime calls and memory ownership behind internal/native.
- Covers timeout inheritance, explicit zero overrides, shared-session preservation, and retry recovery.

The go-binding-automation follow-up contains the delivery layer: embedded native libraries, extraction and cache handling, generated C header and ABI layout verification, cbindgen, lint configuration, CI checks, release assembly, and published-module packaging. GitHub does not support first-class stacked PRs across forks, so that follow-up PR will be opened after this PR merges.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2040,6 +2040,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "tui-test-go"
+version = "0.1.0-beta.3"
+dependencies = [
+ "tui-test-rs",
+]
+
+[[package]]
 name = "tui-test-node"
 version = "0.1.0-beta.3"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,6 @@
 [workspace]
 members = [
+    "bindings/go/native",
     "bindings/js",
     "bindings/python/native",
     "crates/tui-test",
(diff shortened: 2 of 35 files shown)


meza:
@microsoft-github-policy-service agree

cpendery:
Appreciate the work to add go support to tui-test, a few small notes
1) I'd like to hold off on adding go support until we get the 0.1.0 release / features complete
2) Once we are ready to start adding go support, let's split the pr because 7k lines is just too big. In PR 1, let's add the support without any CI / packaging changes, and those can get added in a follow up
3) I think purego is an okay option, but maintaining the abi might be some extra effort. Why'd you decide on it vs. using a different library that requires cgo

meza:
Hey, sounds good! I'll strip out just the go changes and put the ci/packaging in a stacked one for later.

As for purego:

I originally used cgo and the first implementation worked, but it made the binding difficult to install and use.

Users needed cgo enabled, a C compiler, the correct Rust library, build flags and runtime library settings. This meant that `go get` was not enough. It also made cross-compilation and deployment more complicated.

I changed the binding to use `purego` so users only need to run:

```sh
go get github.com/microsoft/tui-test/bindings/go
```

The Go module now includes the compiled Rust libraries. At runtime, the binding selects the correct library, extracts it to the user’s cache, checks its contents and loads it automatically.

meza:
@cpendery I've stripped it back, go the packaging/ci ready on another branch but I can't do stacked prs from forks, so that will have to happen if this gets in at one point. No rush, I can still use this locally for my own project in the meantime

Issue #207, opened 2026-09-05
Named session close blocks behind pending waits


meza:
Closing a named in-process session blocks behind a pending wait instead of interrupting it. `close_all()` interrupts the same wait promptly.

Observed on Windows at commit `2d5f8020ba6eb77b9c4da38892c2f6f200311b90`, while testing a new language binding against the existing Rust runtime.

### Reproduction

1. Start a named session with a program that stays alive.
2. Start a text-locator wait for text that will never appear, with a 5-second timeout.
3. While that wait is pending, close the same named session from another thread.
4. Check whether close completes within 2 seconds.

The close operation misses the 2-second deadline and remains blocked behind the wait. Repeating the scenario with `close_all()` completes promptly.

### Expected behavior

Closing one session should interrupt its pending wait and complete cleanup without waiting for the operation’s timeout, consistently with `close_all()`.

### Cause

[`SessionRegistry::execute`](https://github.com/microsoft/tui-test/blob/2d5f8020ba6eb77b9c4da38892c2f6f200311b90/crates/tui-test/src/runtime.rs#L459) holds the per-session generation mutex throughout operation execution, including waits.

Both `SessionRegistry::close()` and `execute(Operation::Close)` acquire that same mutex before reaching cleanup. They therefore cannot interrupt the operation holding it.

`close_all()` takes a different path: it calls `session.interrupt()` before closing sessions.

### Regression coverage

A regression test should establish that the wait has entered the runtime before invoking close, then verify:

- both named-close entrypoints interrupt the wait promptly
- the session is removed and its retained recording remains accessible
- an existing handle can reopen the named session
- unrelated sessions remain open

This is in the shared Rust runtime. JavaScript and Python use the same named-session path, although the reproduction above was exercised through the new binding.

PR #209, opened 2026-09-05
fix(#207): interrupt pending waits when closing named sessions

meza:
Fixing #207 

---

I'm trusting my AI agent on this one, as I'm not too familiar with Rust. 
Please take it with a pinch of salt and look at it.

### DIFF
--- a/crates/tui-test/src/runtime.rs
+++ b/crates/tui-test/src/runtime.rs
@@ -461,6 +461,9 @@ impl SessionRegistry {
         name: &str,
         operation: Operation,
     ) -> Result<OperationResult, TuiTestError> {
+        if matches!(operation, Operation::Close) {
+            return self.close(name).map(|_| OperationResult::Unit);
+        }
         let generation = self.generation(name);
         let _generation = generation
             .lock()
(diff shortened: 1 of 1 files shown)


crates/tui-test/src/runtime.rs
```diff
@@ -505,11 +510,26 @@ impl SessionRegistry {
...
-        let generation = self.generation(name);
-        let _generation = generation
-            .lock()
-            .unwrap_or_else(std::sync::PoisonError::into_inner);
-        self.close_locked(name)
+        let (result, removed) = {
+            let _lifecycle = self
+                .inner
+                .lifecycle
+                .write()
```
cpendery, line 517: maybe instead of holding lifecycle.write() we can replace the per-name `Mutex<()>` a smaller scoped mutex. It can use an enum to represent the various states:

```rust
enum GenerationPhase {
    Idle,
    Running,
    ClosingRunning,
    ClosingIdle,
}

struct Generation {
    phase: Mutex<GenerationPhase>,
    changed: Condvar,
}
```

PR #211, opened 2026-09-08
feat(terminal): report terminal mode state

aymanbagabas:
`state` now reports which terminal modes the child has turned on.

## Why

Nothing exposed them. `bracketed_paste_mode` existed for the monitor's own use, and every other mode a program switched on was invisible to a test — you could not assert that a TUI entered the alternate screen, enabled bracketed paste, or turned on focus reporting.

## Is it unanimous?

Yes, for a specific set — which is the whole question, so I checked it rather than assumed it.

| backend | how modes are read |
| --- | --- |
| alacritty | `TermMode` bitflags |
| rio | `Mode` bitflags (a fork of alacritty's) |
| ghostty | `Terminal::mode(Mode)`, ~40 modes |
| xterm.js | `term.modes`, a **fixed set of 10** with no "read mode N" call |

xterm.js is the binding constraint, so the vocabulary is a closed enum rather than a mode number. **Nine modes qualify:**

| Variant | Sequence |
| --- | --- |
| `ApplicationCursorKeys` | `CSI ?1 h` |
| `ApplicationKeypad` | `ESC =` |
| `Origin` | `CSI ?6 h` |
| `Wraparound` | `CSI ?7 h` (on by default) |
| `Insert` | `CSI 4 h` |
| `FocusEvents` | `CSI ?1004 h` |
| `BracketedPaste` | `CSI ?2004 h` |
| `AlternateScreen` | `CSI ?1049 h` |
| `CursorVisible` | `CSI ?25 h` (on by default) |

**Deliberately excluded**, because reporting `false` for a mode a backend cannot see would be a wrong answer dressed as a real one:

- reverse wraparound (`?45`) and synchronized output (`?2026`) — ghostty and xterm.js track them, alacritty and rio do not
- `LNM` (`20`) — alacritty, rio and ghostty track it, xterm.js does not

## Two things the conformance suite found, rather than me assuming

**Application keypad is `DECKPAM`, not `CSI ?66 h`.** My first pass used the private-mode form and the case failed on alacritty and rio, which simply do not implement it. `ESC =` / `ESC >` is the form all four honor.

**The alternate screen is not in xterm.js's `modes`.** It comes from `term.buffer.active.type === 'alternate'`, which reports `alternate` exactly while `?1049` or `?47` is in effect. On ghostty it maps to `ALT_SCREEN_SAVE`, the mode `?1049` itself sets, rather than the legacy `?47`.

## Shape

`Emulator::mode` is **required, not defaulted**. A defaulted `false` would make a backend that forgot to implement it look like a terminal where nothing is ever enabled, and no test would catch it — which is exactly how ghostty's `keyboard_mode` went missing for as long as it did. `bracketed_paste_mode` now delegates to it, leaving one implementation per backend instead of two.

`state.modes` is a map with every key always present, so `false` means off rather than "this build does not know that mode".

## Tests

Two conformance cases, so all four backends are held to the same contract and the enum stays a claim rather than a wish list:

- `conformance_terminal_modes_turn_on_and_off` — each mode starts at its documented default, turns on, and turns off.
- `conformance_terminal_modes_are_independent` — setting one mode leaves the other seven at their defaults, so a backend that mapped two variants onto one flag fails here instead of quietly reporting the wrong one.

Verified against a live session too:

```
$ printf '\033[?1h\033[?2004h\033[?1049h' | ...
{ "alternate_screen": true, "application_cursor_keys": true,
  "application_keypad": false, "bracketed_paste": true,
  "focus_events": false, "insert": false, "origin": false,
  "wraparound": true }

$ printf '\033=\033[?7l' | ...
application_keypad = true   wraparound = false
```

`cargo test --workspace --features tui-test-rs/ghostty,tui-test-rs/rio,tui-test-rs/xtermjs` passes, and both `clippy -D warnings` invocations plus `cargo fmt --check` are clean.

## Cursor

`DECTCEM` qualifies from a different direction than the rest: it is not in xterm.js's `modes`, but every backend already had to answer it for the renderer, so it is reported from the same place `cursor_visible` always was. `cursor_visible` now delegates to `mode`, so each backend answers in one place instead of two — rio still reads a hidden cursor as a cursor *shape*, and ghostty now reads `DECTCEM` directly rather than the render snapshot.

The cursor grows the rest of its state. `state.cursor` and `get cursor` report `visible`, `shape`, and `color` beside `x` and `y`.

## CLI

```sh
tui-test get modes                                   # every mode, as JSON
tui-test get cursor --json                           # x, y, visible, shape, color
tui-test expect mode alternate_screen                # on
tui-test expect mode bracketed_paste --off
tui-test expect cursor --hidden
tui-test expect cursor --visible --shape bar
tui-test expect cursor --x 4 --y 0
```

`expect cursor` is a command rather than sugar for `expect mode cursor_visible`. The cursor has four properties and three of them have nowhere else to live, so it earns a cohesive noun and visibility comes along with it. There is deliberately **no** `expect alt-screen`: it would be a second spelling of a single boolean, and an argument for eight more like it.

`expect cursor` checks only the properties named, so asserting a shape leaves position and visibility alone. `--visible` and `--hidden` are separate flags rather than one optional boolean, so naming neither means "do not care" rather than "must be hidden".

An unknown mode is a usage error that names the set:

```
$ tui-test expect mode nonsense
unknown terminal mode 'nonsense'; expected one of: application_cursor_keys,
application_keypad, origin, wraparound, insert, focus_events,
bracketed_paste, alternate_screen, cursor_visible
```

`OperationResult::State` is now boxed: `modes` pushed it far enough past the other variants to trip `large_enum_variant`, and it is returned from every operation.

## Verified live

```
$ printf '\033[?25l\033[?1049h\033[5 q'
get cursor  -> {"x":0,"y":0,"visible":false,"shape":"bar","color":"#c0c0c0"}
get modes   -> alternate_screen=true, cursor_visible=false, wraparound=true, rest false

expect mode alternate_screen          exit 0
expect mode bracketed_paste           exit 1
expect mode bracketed_paste --off     exit 0
expect cursor --hidden --shape bar    exit 0
expect cursor --visible               exit 1
```

## Not in this PR

A locator/`get_by_style` surface for modes, and JS/Python `expect_mode` helpers — the bindings get `state.modes` and the richer `cursor`, but no dedicated assertion methods yet.

### DIFF
--- a/crates/tui-test/src/terminal/ghostty/core.rs
+++ b/crates/tui-test/src/terminal/ghostty/core.rs
@@ -14,7 +14,7 @@ use ghostty_vt::key::{
     Mods as GhosttyMods, OptionAsAlt,
 };
 use ghostty_vt::render::{CellIterator, CursorVisualStyle, RowIterator};
-use ghostty_vt::screen::{Cell as GhosttyCell, CellContentTag, CellWide, GridRef};
+use ghostty_vt::screen::{Cell as GhosttyCell, CellContentTag, CellWide, GridRef, Screen};
 use ghostty_vt::style::{Palette, PaletteIndex, RgbColor, Style, StyleColor, Underline};
 use ghostty_vt::terminal::{
     ConformanceLevel, DeviceAttributeFeature, DeviceAttributes, DeviceType, Mode, Point,
@@ -271,26 +271,24 @@ impl GhosttyCore {
     }
 
     pub(super) fn mode(&self, mode: TerminalMode) -> Result<bool> {
-        // Ghostty honors all three ways into the alternate screen and tracks
-        // each under its own flag: `?47`, `?1047` and `?1049`. The question
-        // `alternate_screen` asks is whether the alternate screen is showing,
-        // so any of them answers it — reading only `?1049` reported "primary"
-        // while ghostty was plainly displaying the alternate buffer.
+        // Ghostty honors all three ways onto the alternate screen — `?47`,
+        // `?1047` and `?1049` — and they share one screen: each dispatches to
+        // `switchScreenMode`, which moves the same `screens.active_key`. The
+        // *mode bits* are what is separate. `setMode` sets one bit per mode
+        // before dispatching, and nothing syncs them, so `?47h` followed by
+        // `?1049l` leaves bit 47 set while the primary screen is showing.
+        //
+        // So no mode bit answers "which screen is showing": reading `?1049`
+        // alone missed the other two entries, and reading all three claimed
+        // the alternate screen after `?1049l` had left it. Ghostty tracks the
+        // active screen directly, which is the question being asked.
         if mode == TerminalMode::AlternateScreen {
-            for flag in [
-                Mode::ALT_SCREEN_SAVE,
-                Mode::ALT_SCREEN,
-                Mode::ALT_SCREEN_LEGACY,
-            ] {
-                if self
-                    .terminal
-                    .mode(flag)
-                    .with_context(|| format!("reading {} mode", mode.name()))?
-                {
-                    return Ok(true);
-                }
-            }
-            return Ok(false);
+            return Ok(matches!(
+                self.terminal
+                    .active_screen()
+                    .context("reading active screen")?,
+                Screen::Alternate
+            ));
         }
         let ghostty_mode = match mode {
             TerminalMode::ApplicationCursorKeys => Mode::DECCKM,
--- a/crates/tui-test/src/terminal/ghostty/core.rs
+++ b/crates/tui-test/src/terminal/ghostty/core.rs
@@ -271,9 +271,27 @@ impl GhosttyCore {
     }
 
     pub(super) fn mode(&self, mode: TerminalMode) -> Result<bool> {
-        // Ghostty tracks `?1049` as the pair it is: a screen swap plus a
-        // cursor save. `ALT_SCREEN_SAVE` is the one `?1049` itself sets, and
-        // it is what a child that sent `?1049` gets told about.
+        // Ghostty honors all three ways into the alternate screen and tracks
+        // each under its own flag: `?47`, `?1047` and `?1049`. The question
+        // `alternate_screen` asks is whether the alternate screen is showing,
+        // so any of them answers it — reading only `?1049` reported "primary"
+        // while ghostty was plainly displaying the alternate buffer.
+        if mode == TerminalMode::AlternateScreen {
+            for flag in [
+                Mode::ALT_SCREEN_SAVE,
+                Mode::ALT_SCREEN,
+                Mode::ALT_SCREEN_LEGACY,
+            ] {
+                if self
+                    .terminal
+                    .mode(flag)
+                    .with_context(|| format!("reading {} mode", mode.name()))?
+                {
+                    return Ok(true);
+                }
+            }
+            return Ok(false);
+        }
         let ghostty_mode = match mode {
             TerminalMode::ApplicationCursorKeys => Mode::DECCKM,
             TerminalMode::ApplicationKeypad => Mode::KEYPAD_KEYS,
@@ -282,7 +300,7 @@ impl GhosttyCore {
             TerminalMode::Insert => Mode::INSERT,
             TerminalMode::FocusEvents => Mode::FOCUS_EVENT,
             TerminalMode::BracketedPaste => Mode::BRACKETED_PASTE,
-            TerminalMode::AlternateScreen => Mode::ALT_SCREEN_SAVE,
+            TerminalMode::AlternateScreen => unreachable!("handled above"),
             TerminalMode::CursorVisible => Mode::CURSOR_VISIBLE,
         };
         self.terminal
--- a/crates/tui-test/src/terminal/ghostty/core.rs
+++ b/crates/tui-test/src/terminal/ghostty/core.rs
@@ -178,7 +178,6 @@ fn cell_from_grid(grid: &GridRef<'_>, links: &mut LinkCache) -> Result<EmuCell> 
 pub(super) struct Frame {
     pub(super) rows: Vec<Vec<EmuCell>>,
     pub(super) cursor: (u16, u16),
-    pub(super) cursor_visible: bool,
     pub(super) cursor_shape: CursorShape,
 }
 
@@ -284,6 +283,7 @@ impl GhosttyCore {
             TerminalMode::FocusEvents => Mode::FOCUS_EVENT,
             TerminalMode::BracketedPaste => Mode::BRACKETED_PASTE,
             TerminalMode::AlternateScreen => Mode::ALT_SCREEN_SAVE,
+            TerminalMode::CursorVisible => Mode::CURSOR_VISIBLE,
         };
         self.terminal
             .mode(ghostty_mode)
@@ -481,9 +481,6 @@ impl GhosttyCore {
                 .context("reading cursor row")?
                 .min(rows.saturating_sub(1)),
         );
-        let cursor_visible = snapshot
-            .cursor_visible()
-            .context("reading cursor visibility")?;
         let cursor_shape = match snapshot
             .cursor_visual_style()
             .context("reading cursor shape")?
@@ -565,7 +562,6 @@ impl GhosttyCore {
         Ok(Frame {
             rows: output,
             cursor,
-            cursor_visible,
             cursor_shape,
         })
     }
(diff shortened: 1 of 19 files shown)


crates/tui-test/src/terminal/ghostty/core.rs
```diff
@@ -228,10 +227,24 @@ impl GhosttyCore {
...
+        // it is what a child that sent `?1049` gets told about.
+        let ghostty_mode = match mode {
+            TerminalMode::ApplicationCursorKeys => Mode::DECCKM,
+            TerminalMode::ApplicationKeypad => Mode::KEYPAD_KEYS,
+            TerminalMode::Origin => Mode::ORIGIN,
+            TerminalMode::Wraparound => Mode::WRAPAROUND,
+            TerminalMode::Insert => Mode::INSERT,
+            TerminalMode::FocusEvents => Mode::FOCUS_EVENT,
+            TerminalMode::BracketedPaste => Mode::BRACKETED_PASTE,
+            TerminalMode::AlternateScreen => Mode::ALT_SCREEN_SAVE,
```
cpendery: While this uses the `CSI ? 1049` flag, I'm pretty sure `CSI ? 47 h` and `CSI ? 1047 h` also open the alt buffer for Ghostty. Rio/Alacritty ignore those entirely

crates/tui-test/src/terminal/rio.rs
cpendery, line 433: I don't think we want rio's hidden cursor to be reported as block for the metadata

cpendery (commented):
Overall looks good. Double check me on this but won't `CSI ? 1000 h` alone enable click tracking but our new public field says `"none"`

crates/tui-test/src/terminal/rio.rs
aymanbagabas: that's just how I believe rio keeps track of its `CSI ? 25 l` state (cursor hidden)

crates/tui-test/src/terminal/ghostty/core.rs
aymanbagabas: that's right, perhaps we should have `AlternateScreenSaveCursor` as separate mode

crates/tui-test/src/terminal/rio.rs
aymanbagabas: Right, and it was a real bug — fixed in 3fb9cdd.

Rio keeps the shape in `Term.cursor_shape`, but `Term::cursor()` masks the derived `content` to `Hidden` whenever the cursor is not drawn — hidden by DECTCEM, and also when the view is scrolled back. So reading `cursor().content` turned every hidden cursor into a block and lost the shape the child had asked for. Reading `cursor_shape` instead gives the unmasked value, and whether the cursor is drawn is already reported by `CursorVisible`, so the two were answering the same question and disagreeing.

Confirmed with a conformance test (set bar, hide, show): alacritty, ghostty and xterm.js all keep the shape, and rio was the only one that did not. It passes on all four now, so this needed no divergence.

crates/tui-test/src/terminal/ghostty/core.rs
aymanbagabas: Confirmed, and there was a bug behind it. Fixed in f1372d3.

I probed all three sequences on all four backends (write `primary`, send the sequence, check whether the text is still visible against what the flag says):

| | `?47h` | `?1047h` | `?1049h` |
|---|---|---|---|
| alacritty | ignored | ignored | switches, flag ✓ |
| rio | ignored | ignored | switches, flag ✓ |
| ghostty | **switches, flag ✗** | **switches, flag ✗** | switches, flag ✓ |
| xterm.js | switches, flag ✓ | switches, flag ✓ | switches, flag ✓ |

So you were right about ghostty, and xterm.js honors them too. The bug is the ✗ column: ghostty tracks each spelling under its own mode (`ALT_SCREEN_LEGACY`/`ALT_SCREEN`/`ALT_SCREEN_SAVE`), and we read only `?1049`, so `alternate_screen` reported "primary" while ghostty was displaying the alternate buffer. It now reads all three, and the flag matches the visible screen everywhere.

On @aymanbagabas's suggestion of a separate `AlternateScreenSaveCursor` — I'd argue against it, for two reasons. It fails this PR's own bar that a mode be unanimously supported, since alacritty and rio ignore `?47`/`?1047` outright. And `?1049` versus `?47` isn't two states a terminal can be in, it's two ways into the same one plus a cursor save; a program that sent `?1049h` would then have to report true for both or true for one, and neither tells a test author anything. The cursor-save half is `?1048`, which is a save/restore action rather than a mode you can read back.

Instead there's a conformance test pinning the flag to the visible screen for all three, and `NoLegacyAlternateScreen` declared for alacritty and rio so the difference is recorded rather than assumed. Nothing real depends on it: every backend honors `?1049`, which is what a full-screen program sends.

aymanbagabas:
> Double check me on this but won't `CSI ? 1000 h` alone enable click tracking but our new public field says `none`

You're right. Fixed in ce37293.

`MouseModeTracker::mode()` was gated on the child *also* having sent `CSI ?1006 h`, so `?1000h` on its own reported `none` for a program that was plainly reading clicks. Tracking level and report encoding are separate axes, and the field was conflating them.

The gate itself is legitimate, just not here: it belongs to the monitor, which mirrors the child's tracking onto the watching terminal and forwards what comes back. It can only encode SGR, so relaying to a child that asked for the legacy `CSI M` form would feed it reports it cannot parse. That is now `relayable()`, and `mode()` answers what the child asked for.

Two tests pin the split: one that `?1000h` alone is `click` whatever the encoding, and one that `?1016h` (pixel coordinates) makes tracking unrelayable while the child is still tracking drags.

Also in this push, both from your inline comments — the rio hidden-cursor shape and the ghostty alternate-screen flag were each real bugs. Details in those threads.

crates/tui-test/src/terminal/ghostty/core.rs
aymanbagabas: Correction to my reply above — @aymanbagabas was right that the three share one screen, and my first fix was wrong because of it. Now fixed properly in 05da9e8.

Reading ghostty's source rather than inferring from behavior:

- **The screen is shared.** `alt_screen_legacy` (47), `alt_screen` (1047) and `alt_screen_save_cursor_clear_enter` (1049) all dispatch to `Terminal.switchScreenMode`, which calls `switchScreen(.alternate/.primary)` on the one `screens.active_key` (`stream_terminal.zig:1580-1582`, `Terminal.zig:4826`).
- **The mode bits are not.** `setMode` does `self.terminal.modes.set(mode, enabled)` generically *before* dispatching (`stream_terminal.zig:1563-1565`), one bit per mode, and nothing syncs them.

So `?47h` then `?1049l` leaves bit 47 set with the primary screen showing. My first fix ORed the three bits, which fixed entering and broke leaving — it would have reported the alternate screen after a program had left it. The conformance test missed it because it only covered entering.

It now reads `Terminal::active_screen()`, which is the question being asked, and the test covers leaving for all three spellings. Verified it fails against the OR version.

Doc surfaces updated too: the `TerminalMode::AlternateScreen` comment, the reference table, and `set_sequence`/`reset_sequence` all said `?1049` was *the* sequence rather than one of three.

aymanbagabas: Leaving this one open rather than resolving it, since it still holds an open design question: @aymanbagabas suggested a separate `AlternateScreenSaveCursor` mode and I argued against it, so it should not be closed on my say-so.

The bug part is fixed and verified either way. The argument against splitting is now stronger than when I first made it: the three spellings share a single screen in ghostty, so there is only one screen state to report, and a second mode would have nothing distinct to say. What differs is only the mode bits, and those turned out to be the wrong thing to read at all.

Happy to add it if you still want it — say the word and I will.

cpendery: Could we just use `Terminal::active_screen()` instead of combing the mode flags?

aymanbagabas: That is exactly what it does now — 05da9e8, currently the tip of the branch:

```rust
if mode == TerminalMode::AlternateScreen {
    return Ok(matches!(
        self.terminal.active_screen().context("reading active screen")?,
        Screen::Alternate
    ));
}
```

`grep ALT_SCREEN` across the crate now returns nothing, so no mode flags are combed for this at all.

GitHub is anchoring this thread to the original diff hunk, which still shows the `TerminalMode::AlternateScreen => Mode::ALT_SCREEN_SAVE` line — that is probably what you were reading. Sorry, my correction two comments up buried the actual change under the reasoning.

Short version: ORing the three bits was my first attempt and it was wrong, because `?1049l` clears only its own bit while restoring the primary screen, so it would have claimed the alternate screen after a program had left it. `active_screen()` is the right question, as you say.

cpendery (dismissed):

cpendery (dismissed):

cpendery (dismissed):

cpendery (approved):

PR #233, opened 2026-09-14
test: cover standalone trace reports with Playwright

cpendery:


cpendery:
Consolidated into #232, which now contains all three remaining reviewed commits against main. No file changes were omitted.

